PR #7, opened 2022-12-02
ENH: Add my "MPFR" example

seberg:
Admittedly, it really needs some tests and I just copied the meson mostly (with adding the `mpfr` dependency.

Might be interesting since I do use some C++ templating, use with care, those should be compiled in a way that no C++ exceptions can happen, etc. but I do not think that is the case.


ngoldbaum:
We can iterate later on adding tests but let's get this in to serve as an example.

PR #8, opened 2022-12-03
ENH: Add `MPFloat.prec` and `MPFDType.prec`, also add basic scalar tests

seberg:
Adds the `.prec` attributes, since it wasn't actually reachable before (only visible in the repr).

Plus, started writing some tests for scalars.


ngoldbaum:
If you’d like you can add a step to the github actions workflow so `mpfdtype` gets compiled and the tests run on CI.

PR #10, opened 2022-12-06
Add asciidtype

ngoldbaum:
Adds a fixed-width ASCII dtype. This is still pretty bare-bones but I figured it was worth getting in the repo in case anyone wants to look at a string dtype. I'll be continuing to work on this.

Going to leave this open until tomorrow for reviews and then self-merge then if I don't get any reviews, assuming tests pass.

### DIFF
--- a/asciidtype/asciidtype/src/casts.c
+++ b/asciidtype/asciidtype/src/casts.c
@@ -62,46 +62,11 @@ ascii_to_ascii_contiguous(PyArrayMethod_Context *context, char *const data[],
 }
 
 static int
-ascii_to_ascii_strided(PyArrayMethod_Context *context, char *const data[],
-                       npy_intp const dimensions[], npy_intp const strides[],
-                       NpyAuxData *NPY_UNUSED(auxdata))
-{
-    PyArray_Descr **descrs = context->descriptors;
-    long in_size = ((ASCIIDTypeObject *)descrs[0])->size;
-    long out_size = ((ASCIIDTypeObject *)descrs[1])->size;
-    long copy_size;
-
-    if (out_size > in_size) {
-        copy_size = in_size;
-    }
-    else {
-        copy_size = out_size;
-    }
-
-    npy_intp N = dimensions[0];
-    char *in = data[0];
-    char *out = data[1];
-    npy_intp in_stride = strides[0];
-    npy_intp out_stride = strides[1];
-
-    while (N--) {
-        for (int i = 0; i < copy_size; i++) {
-            *(out + i) = *(in + i);
-        }
-        for (int i = copy_size; i < out_size; i++) {
-            *(out + i) = '\0';
-        }
-        in += in_stride;
-        out += out_stride;
-    }
-
-    return 0;
-}
-
-static int
-ascii_to_ascii_unaligned(PyArrayMethod_Context *context, char *const data[],
-                         npy_intp const dimensions[], npy_intp const strides[],
-                         NpyAuxData *NPY_UNUSED(auxdata))
+ascii_to_ascii_strided_or_unaligned(PyArrayMethod_Context *context,
+                                    char *const data[],
+                                    npy_intp const dimensions[],
+                                    npy_intp const strides[],
+                                    NpyAuxData *NPY_UNUSED(auxdata))
 {
     PyArray_Descr **descrs = context->descriptors;
     long in_size = ((ASCIIDTypeObject *)descrs[0])->size;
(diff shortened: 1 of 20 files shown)


asciidtype/asciidtype/src/casts.c
```diff
@@ -0,0 +1,180 @@
...
+                        PyArrayMethod_StridedLoop **out_loop,
+                        NpyAuxData **NPY_UNUSED(out_transferdata),
+                        NPY_ARRAYMETHOD_FLAGS *flags)
+{
+    PyArray_Descr **descrs = context->descriptors;
+
+    int contig = (strides[0] == ((ASCIIDTypeObject *)descrs[0])->size *
+                                        sizeof(char) &&
+                  strides[1] == ((ASCIIDTypeObject *)descrs[1])->size *
+                                        sizeof(char));
```
seberg, line 149: It seems to me that your contig loop also assumes `descrs[0]->size == descrs[1]->size` which is not checked below?

asciidtype/asciidtype/src/casts.c
```diff
@@ -0,0 +1,180 @@
...
+
+    npy_intp N = dimensions[0] * size;
+    char *in = data[0];
+    char *out = data[1];
+
+    while (N--) {
+        *out = *in;
+        out++;
+        in++;
+    }
```
seberg, line 59: could even use plain `memcpy` if you like.

asciidtype/asciidtype/src/casts.c
```diff
@@ -0,0 +1,180 @@
...
+                                        sizeof(char));
+
+    if (aligned && contig) {
+        *out_loop = (PyArrayMethod_StridedLoop *)&ascii_to_ascii_contiguous;
+    }
+    else if (aligned) {
+        *out_loop = (PyArrayMethod_StridedLoop *)&ascii_to_ascii_strided;
+    }
+    else {
+        *out_loop = (PyArrayMethod_StridedLoop *)&ascii_to_ascii_unaligned;
```
seberg, line 158: The aligment of `char` will be 1, so while I think I currently sometimes pass "unaligned" to not bother figuring out alignment, you could just use the same loop for both.  I suppose you can also argue that char isn't standardized to 1 byte :).

asciidtype/asciidtype/src/dtype.c
```diff
@@ -0,0 +1,265 @@
...
+
+/*
+ * This is used to determine the correct dtype to return when operations mix
+ * dtypes (I think?). For now just return the first one.
+ */
+static ASCIIDTypeObject *
+common_instance(ASCIIDTypeObject *dtype1, ASCIIDTypeObject *dtype2)
+{
+    if (!PyObject_RichCompareBool((PyObject *)dtype1, (PyObject *)dtype2,
+                                  Py_EQ)) {
```
seberg, line 67: N.B.:  This is actually the same as checking whether your resolver returns `NO_CASTING` or `EQUIV_CASTING`.

seberg (commented):
Just a brief look through if I saw anything.  I wonder if it would make sense to make the scalar just a subclass of `str` (i.e. on the path to having no scalar).
The main problem with that is NumPy's insistence to convert 0-D arrays to scalars  (so long we enforce getitem not having a scalar is probably even fine).

asciidtype/tests/test_asciidtype.py
```diff
@@ -0,0 +1,20 @@
...
+def test_scalar_creation():
+    dtype = ASCIIDType(7)
+    ASCIIScalar('string', dtype)
+
+
+def test_creation_with_explicit_dtype():
+    dtype = ASCIIDType(7)
+    arr = np.array(["hello", "this", "is", "an", "array"], dtype=dtype)
+    assert repr(arr) == (
+        "array(['hello', 'this', 'is', 'an', 'array'], dtype=ASCIIDType(7))")
```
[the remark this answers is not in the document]
seberg: I was going to say yes, but currently the `dtype=ASCIIDType` isn't passed through correctly, it probably works for `np.core.multiarray._multiarray_umath._discover_array_parameters()`, though.

Passing through the DType type/class rather than just the dtype is something we/I need to look into...  (A similar alternative might be to introduce abstract instances, that cannot be attached to an array, but represent the same thing as ASCIIDType).

asciidtype/asciidtype/src/casts.c
seberg: Its hard, but there are four paths right now how you can reach unaligned casts:
* I think I had some scalar paths where I didn't bother checking alignment and just calling the unaligned version).  (I have to double check this, but I think it happens for casts between dtypes when assigning a single value.)
* You can manage it by manually ensuring you have unaligned data ;).  But that is tricky, e.g. with `np.zeros(1001)[1:].view("U10")`.
* Structured dtypes by default use packed structs rather than aligned ones.  So if you do `dtype=[("a", "?"), ("b", "U5")]` than the `"b"` field is unaligned.
* Structured dtypes are a mess and they (currently) do not even try to pass on correct alignment to the field dtypes when necessary (i.e. in casts).  That is something that can be done, but I really would like to pass through the *actual* alignment, rather than just "its all aligned".  (https://github.com/numpy/numpy/issues/17359)

The last point is the main one really, the second one is rather theoretical...  Which is why my suggestion to allow flagging "do not allow unaligned".

ngoldbaum: Thanks! Having such a flag, with the default to not allow aunaligned, would certainly reduce the conceptual overhead quite a bit.

asciidtype/tests/test_asciidtype.py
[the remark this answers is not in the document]
ngoldbaum: I opened https://github.com/numpy/numpy/issues/22756 to track this, seems like it would be useful.

PR #11, opened 2022-12-07
Fix bugs around null termination bytes

ngoldbaum:
I wasn't accounting properly for null termination bytes in my original implementation. I now append a null byte to the result of getitem.

I've also added tests and fixed some bugs around making sure data are truncated properly at array creation time and casting time if the dtype is too narrow to contain the full input data.

### DIFF
--- a/asciidtype/tests/test_asciidtype.py
+++ b/asciidtype/tests/test_asciidtype.py
@@ -41,12 +41,12 @@ def test_creation_truncation():
     assert repr(arr) == (
         "array(['h', 't', 'i', 'a', 'a'], dtype=ASCIIDType(1))"
     )
-    assert arr.tobytes() == b"h\x00t\x00i\x00a\x00a\x00"
+    assert arr.tobytes() == b"htiaa"
 
-    dtype = ASCIIDType()
-    arr = np.array(["hello", "this", "is", "an", "array"], dtype=dtype)
-    assert repr(arr) == ("array(['', '', '', '', ''], dtype=ASCIIDType(0))")
-    assert arr.tobytes() == b"\x00\x00\x00\x00\x00"
+    # dtype = ASCIIDType()
+    # arr = np.array(["hello", "this", "is", "an", "array"], dtype=dtype)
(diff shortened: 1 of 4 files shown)


seberg:
Just to note, NumPy's string/bytes does not store that null terminator, there is an implicit null terminator behind things.  Of course you have to be very careful to not copy it.

mattip:
I think this is wrong: the dtype itemsize should match the name.

asciidtype/tests/test_asciidtype.py
```diff
@@ -10,11 +10,64 @@ def test_dtype_creation():
...
+    arr = np.array(inp, dtype=dtype)
+    assert repr(arr) == (
+        "array(['h', 't', 'i', 'a', 'a'], dtype=ASCIIDType(1))"
+    )
+    assert arr.tobytes() == b"htiaa"
+
+    # dtype = ASCIIDType()
+    # arr = np.array(["hello", "this", "is", "an", "array"], dtype=dtype)
+    # assert repr(arr) == ("array(['', '', '', '', ''], dtype=ASCIIDType(0))")
+    # assert arr.tobytes() == b""
```
ngoldbaum, line 49: These tests are commented out because they depend on https://github.com/numpy/numpy/pull/22763 being merged in numpy.

ngoldbaum:
Now it doesn’t store the null characters anymore and `getitem` will pad the result with null.

I also realized the contiguous casting loop isn’t terribly useful because almost all casts aren’t aligned, so I’ve made it so there’s only one casting loop.

ngoldbaum:
> because almost all casts aren’t aligned

In particular, I found that the check for whether a cast is aligned does `npy_uint_alignment(dtype->elsize)`, which only returns true dtypes that are 1, 4, 8, or 16 bytes wide.

seberg:
Eeek, the copying code has additional alignment for the purpose of complex loops...  Because we copy complex64 (two 32bit floats) via uint64 which has a lrger alignment.

None of that is ideal, but due to that the alignment passed in is larger than you would think.

PR #12, opened 2022-12-12
Fix quaddtype debug symbols

peytondmurray:


quaddtype/quaddtype/src/umath.c
```diff
@@ -56,9 +56,18 @@ quad_multiply_resolve_descriptors(PyObject *self, PyArray_DTypeMeta *dtypes[],
...
+        loop_descrs[1] = given_descrs[0];
+        return NPY_NO_CASTING;
+    }
+    Py_INCREF(given_descrs[1]);
+    loop_descrs[1] = given_descrs[1];
+
     Py_INCREF(given_descrs[1]);
     loop_descrs[1] = given_descrs[1];
-    return NPY_NO_CASTING;
+    return NPY_SAFE_CASTING;
```
seberg, line 70: You know that you don't have any byte-swaps or so, so this is always NO_CASTING.  In principle, you could always return the singleton instance also.  (The only reason NumPy has more instances is byte-swapping and metadata, which you don't have to support.)

Bu it's fine to just delete it for this DType really.

The function currently doesn't fill in the output `loop_descrs[2]`?

ngoldbaum:
If you'd like, please feel free to add some tests for `quaddtype` and then hook that up to the github actions workflow. I merged this quickly (before Sebastian had time to comment..) but if that's not what you want please let me know.

peytondmurray:
No problem, and thanks for the feedback to you both. I'm still filling out the implementation, so I'll definitely add tests for this now.

PR #13, opened 2022-12-12
Add round-trip casts between unicode and ASCIIDType

ngoldbaum:
Since Numpy stores unicode data internally as UCS4, I can do these casts relatively straightforwardly without worrying about encoding details. 

For unicode to ASCII, I check character by character (e.g. 4 bytes at a time) if the UCS4 character is valid ASCII, if it is I assign the first byte of the character to the corresponding byte of the output array. If we find an invalid ASCII character, we re-acquire the GIL, set a `TypeError`, release the GIL, and return.

For ASCII to unicode, no error checking is needed, so we just set the ASCII character to the first byte of the corresponding character in the output array, and set the rest of the bytes in the character to zero.

This also includes some other misc fixes I noticed along the way around error checking and handling reference counts.

Finally, mostly as a note to myself to check this tomorrow, I noticed that if I do a round-trip from unicode to ascii back to unicode, but I phrase it like this:

```
arr = np.array(["hello", "this", "is", "an", "array"])
ascii_arr = arr.astype(ASCIIDType(5))
round_trip_arr = ascii_arr.astype(np.unicode_)
```

the resulting array will still have `ASCIIDType(5)` as the dtype. If I specify the output dtype as a string (`ascii_arr.astype('U5')` as I've done in the round-trip test I added in this PR), it works fine. I need to look into the implementation of `astype` to understand why that's happening, I suspect it's a numpy bug.

### DIFF
--- a/asciidtype/asciidtype/src/casts.c
+++ b/asciidtype/asciidtype/src/casts.c
@@ -97,8 +97,11 @@ unicode_to_ascii_resolve_descriptors(PyObject *NPY_UNUSED(self),
     Py_INCREF(given_descrs[0]);
     loop_descrs[0] = given_descrs[0];
     if (given_descrs[1] == NULL) {
-        Py_INCREF(given_descrs[0]);
-        loop_descrs[1] = given_descrs[0];
+        // numpy stores unicode as UCS4 (4 bytes wide), so bitshift
+        // by 2 to get the number of ASCII bytes needed
+        long size = (loop_descrs[0]->elsize) >> 2;
+        ASCIIDTypeObject *ascii_descr = new_asciidtype_instance(size);
+        loop_descrs[1] = (PyArray_Descr *)ascii_descr;
     }
     else {
         Py_INCREF(given_descrs[1]);
@@ -245,8 +248,12 @@ ascii_to_unicode_resolve_descriptors(PyObject *NPY_UNUSED(self),
     Py_INCREF(given_descrs[0]);
     loop_descrs[0] = given_descrs[0];
     if (given_descrs[1] == NULL) {
-        Py_INCREF(given_descrs[0]);
-        loop_descrs[1] = given_descrs[0];
+        PyArray_Descr *unicode_descr = PyArray_DescrNewFromType(NPY_UNICODE);
+        long num_ascii_bytes = ((ASCIIDTypeObject *)given_descrs[0])->size;
+        // numpy stores unicode as UCS4 (4 bytes wide), so bitshift
+        // by 2 to get the number of bytes needed to store the UCS4 charaters
+        unicode_descr->elsize = num_ascii_bytes << 2;
+        loop_descrs[1] = unicode_descr;
     }
     else {
         Py_INCREF(given_descrs[1]);
--- a/asciidtype/asciidtype/src/casts.c
+++ b/asciidtype/asciidtype/src/casts.c
@@ -28,13 +28,17 @@ ascii_to_ascii_resolve_descriptors(PyObject *NPY_UNUSED(self),
         loop_descrs[1] = given_descrs[1];
     }
 
-    if (((ASCIIDTypeObject *)loop_descrs[0])->size ==
-        ((ASCIIDTypeObject *)loop_descrs[1])->size) {
+    long in_size = ((ASCIIDTypeObject *)loop_descrs[0])->size;
+    long out_size = ((ASCIIDTypeObject *)loop_descrs[1])->size;
+
+    if (in_size == out_size) {
         *view_offset = 0;
         return NPY_NO_CASTING;
     }
-
-    return NPY_SAME_KIND_CASTING;
+    else if (in_size > out_size) {
+        return NPY_UNSAFE_CASTING;
+    }
+    return NPY_SAFE_CASTING;
 }
 
 static int
@@ -72,21 +76,6 @@ ascii_to_ascii(PyArrayMethod_Context *context, char *const data[],
     return 0;
 }
 
-static int
-ascii_to_ascii_get_loop(PyArrayMethod_Context *NPY_UNUSED(context),
-                        int NPY_UNUSED(aligned),
-                        int NPY_UNUSED(move_references),
-                        const npy_intp *NPY_UNUSED(strides),
-                        PyArrayMethod_StridedLoop **out_loop,
-                        NpyAuxData **NPY_UNUSED(out_transferdata),
-                        NPY_ARRAYMETHOD_FLAGS *flags)
-{
-    *out_loop = (PyArrayMethod_StridedLoop *)&ascii_to_ascii;
-
-    *flags = 0;
-    return 0;
-}
-
 static NPY_CASTING
 unicode_to_ascii_resolve_descriptors(PyObject *NPY_UNUSED(self),
                                      PyArray_DTypeMeta *NPY_UNUSED(dtypes[2]),
@@ -96,11 +85,11 @@ unicode_to_ascii_resolve_descriptors(PyObject *NPY_UNUSED(self),
 {
     Py_INCREF(given_descrs[0]);
     loop_descrs[0] = given_descrs[0];
+    // numpy stores unicode as UCS4 (4 bytes wide), so bitshift
+    // by 2 to get the number of ASCII bytes needed
+    long in_size = (loop_descrs[0]->elsize) >> 2;
     if (given_descrs[1] == NULL) {
-        // numpy stores unicode as UCS4 (4 bytes wide), so bitshift
-        // by 2 to get the number of ASCII bytes needed
-        long size = (loop_descrs[0]->elsize) >> 2;
-        ASCIIDTypeObject *ascii_descr = new_asciidtype_instance(size);
+        ASCIIDTypeObject *ascii_descr = new_asciidtype_instance(in_size);
         loop_descrs[1] = (PyArray_Descr *)ascii_descr;
     }
     else {
@@ -108,7 +97,13 @@ unicode_to_ascii_resolve_descriptors(PyObject *NPY_UNUSED(self),
         loop_descrs[1] = given_descrs[1];
     }
 
-    return NPY_SAME_KIND_CASTING;
+    long out_size = ((ASCIIDTypeObject *)loop_descrs[1])->size;
+
+    if (out_size >= in_size) {
+        return NPY_SAFE_CASTING;
+    }
+
+    return NPY_UNSAFE_CASTING;
 }
 
 static int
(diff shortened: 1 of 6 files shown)


asciidtype/asciidtype/src/casts.c
```diff
@@ -86,6 +87,190 @@ ascii_to_ascii_get_loop(PyArrayMethod_Context *context, int aligned,
...
+                                     PyArray_DTypeMeta *NPY_UNUSED(dtypes[2]),
+                                     PyArray_Descr *given_descrs[2],
+                                     PyArray_Descr *loop_descrs[2],
+                                     npy_intp *NPY_UNUSED(view_offset))
+{
+    Py_INCREF(given_descrs[0]);
+    loop_descrs[0] = given_descrs[0];
+    if (given_descrs[1] == NULL) {
+        Py_INCREF(given_descrs[0]);
+        loop_descrs[1] = given_descrs[0];
```
seberg, line 101: This is not correct, I guess it may be a bit hard to reach right now.  `given_descrs[0]` will be a unicode dtype instance here, `loop_descr[1]` must be an ASCII dtype instance.

The correct thing will be to return an ASCII dtype instance.

N.B.: It would even be acceptable to always return an ASCII dtype instance of the same length as the unicode one (NumPy will chain the Unicode -> ASCII and ASCII -> ASCII cast, where the second one would change the size).

asciidtype/asciidtype/src/casts.c
```diff
@@ -86,6 +87,190 @@ ascii_to_ascii_get_loop(PyArrayMethod_Context *context, int aligned,
...
+                          int NPY_UNUSED(aligned),
+                          int NPY_UNUSED(move_references),
+                          const npy_intp *NPY_UNUSED(strides),
+                          PyArrayMethod_StridedLoop **out_loop,
+                          NpyAuxData **NPY_UNUSED(out_transferdata),
+                          NPY_ARRAYMETHOD_FLAGS *flags)
+{
+    *out_loop = (PyArrayMethod_StridedLoop *)&unicode_to_ascii;
+
+    *flags = 0;
```
seberg, line 191: Could actually use `NPY_METH_NO_FLOATINGPOINT_ERRORS`.  Not that it matters much.

asciidtype/asciidtype/src/casts.c
```diff
@@ -102,3 +287,59 @@ PyArrayMethod_Spec ASCIIToASCIICastSpec = {
...
+static PyType_Slot u2a_slots[] = {
+        {NPY_METH_resolve_descriptors, &unicode_to_ascii_resolve_descriptors},
+        {_NPY_METH_get_loop, &unicode_to_ascii_get_loop},
+        {0, NULL}};
+
+static char *u2a_name = "cast_Unicode_to_ASCIIDType";
+
+static PyType_Slot a2u_slots[] = {
+        {NPY_METH_resolve_descriptors, &ascii_to_unicode_resolve_descriptors},
+        {_NPY_METH_get_loop, &ascii_to_unicode_get_loop},
```
seberg, line 300: No need for get_loop, can just use `NPY_METH_strided_loop` directly.  OTOH, maybe it doesn't add a lot of convenience anyways...

asciidtype/tests/test_asciidtype.py
```diff
@@ -50,24 +51,61 @@ def test_creation_truncation():
...
+        ["😀", "hello", "some", "ascii"],
+        ["hello", "some", "ascii", "😀"],
+    ]
+    for inp in inps:
+        arr = np.array(inp)
+        with pytest.raises(
+            TypeError,
+            match="Can only store ASCII text in a ASCIIDType array.",
+        ):
+            arr.astype(ASCIIDType(5))
```
seberg, line 111: I am thinking `astype` already supports `arr.astype(ASCIIDType)`, but I am not sure.  It is definitely reachable e.g. via ufuncs, but that is a bit trickier.

asciidtype/asciidtype/src/casts.c
```diff
@@ -86,6 +87,190 @@ ascii_to_ascii_get_loop(PyArrayMethod_Context *context, int aligned,
...
+    long in_size = ((ASCIIDTypeObject *)descrs[0])->size;
+    long out_size = (descrs[1]->elsize) / 4;
+    long copy_size;
+
+    if (out_size > in_size) {
+        copy_size = in_size;
+    }
+    else {
+        copy_size = out_size;
+    }
```
seberg, line 210: Ah, this type of setup could of course also be done in the `get_loop` if it helps a lot.  But it hardly matters in practice...
(I guess it might for HPy support or so, but that is another tricky thing to figure out one day.)

asciidtype/asciidtype/src/casts.c
```diff
@@ -102,3 +287,59 @@ PyArrayMethod_Spec ASCIIToASCIICastSpec = {
...
+    a2u_dtypes[0] = NULL;
+    a2u_dtypes[1] = &PyArray_UnicodeDType;
+
+    PyArrayMethod_Spec *ASCIIToUnicodeCastSpec =
+            malloc(sizeof(PyArrayMethod_Spec));
+
+    ASCIIToUnicodeCastSpec->name = a2u_name;
+    ASCIIToUnicodeCastSpec->nin = 1;
+    ASCIIToUnicodeCastSpec->nout = 1;
+    ASCIIToUnicodeCastSpec->flags = NPY_METH_SUPPORTS_UNALIGNED;
```
seberg, line 333: You don't need to support unaligned here, numpy will chain the cast if necessary, and it should be relatively rare (and if not, we should probably fix that part and not this here).

asciidtype/asciidtype/src/casts.c
```diff
@@ -86,6 +87,190 @@ ascii_to_ascii_get_loop(PyArrayMethod_Context *context, int aligned,
...
+    npy_intp in_stride = strides[0];
+    npy_intp out_stride = strides[1];
+
+    while (N--) {
+        // copy ASCII input to first byte, fill rest with zeros
+        for (int i = 0; i < copy_size; i++) {
+            *(out + i * 4) = *(in + i);
+            for (int j = 1; j < 4; j++) {
+                *(out + i * 4 + j) = '\0';
+            }
```
seberg, line 224: If you don't bother about supporting unaligned access, you could just write:
```
((Py_UCS4 *)out)[i] = in[i];
```

seberg: Ah, you need to make sure to use `unsigned char` in that case.  But that would also make the check for ASCII much more clear.  (Could also use `Py_UCS1` if you like.)

seberg (commented):
The resolvers are not quite correct and I would suggest to use a type alias for `unsigned char` (maybe just `Py_UCS1`).

Otherwise just a few smaller comments.

asciidtype/asciidtype/src/casts.c
ngoldbaum: Thank you for pointing this out, it was causing the issue I mentioned in the PR description, since `arr.astype(np.unicode_)` is going through `resolve_descriptors` to get the output dtype.

asciidtype/tests/test_asciidtype.py
ngoldbaum: It doesn't support it yet, that's https://github.com/numpy/numpy/issues/22756. I'm going to try poking at that once I finish here.

seberg: Ah, right... I didn't change it anywhere, I guess.  Starting with `astype()` is probably good since it is likely clearer or at least simpler than `np.array()`.

ngoldbaum:
Thanks so much for the comments @seberg, I learn a ton every time you give me code review.

Would you mind taking another look at this when you have a chance?

asciidtype/asciidtype/src/casts.c
```diff
@@ -72,33 +76,243 @@ ascii_to_ascii(PyArrayMethod_Context *context, char *const data[],
...
+
+    PyArrayMethod_Spec *UnicodeToASCIICastSpec =
+            malloc(sizeof(PyArrayMethod_Spec));
+
+    UnicodeToASCIICastSpec->name = u2a_name;
+    UnicodeToASCIICastSpec->nin = 1;
+    UnicodeToASCIICastSpec->nout = 1;
+    UnicodeToASCIICastSpec->casting = NPY_UNSAFE_CASTING,
+    UnicodeToASCIICastSpec->flags =
+            (NPY_METH_NO_FLOATINGPOINT_ERRORS | NPY_METH_REQUIRES_PYAPI);
```
ngoldbaum, line 292: I didn’t know about `NPY_METH_REQUIRES_PYAPI` until I happened to come across it in the experiemental dtype header today. Out of curiosity, what are the downsides of not releasing the GIL in a casting or ufunc loop? For this one I only need a python API function for error handling, so I could re-acquire the GIL manually in the error condition as I originally had it.

seberg: You want to release the GIL as much as possible (except for very small amount of work, since there is a cost to releasing; although maybe that cost got reduced also).

seberg: N.B.: we are relying (right now) on the fact that in CPython (and pypy cpyext) accessing the object is OK even without the GIL so long we only get the `elsize`, etc.
For general Python implementations, this may need to happen in the setup, or we have to think about how the `descrs` are passed exactly.

asciidtype/asciidtype/src/casts.c
```diff
@@ -72,33 +76,243 @@ ascii_to_ascii(PyArrayMethod_Context *context, char *const data[],
...
+                                     npy_intp *NPY_UNUSED(view_offset))
+{
+    Py_INCREF(given_descrs[0]);
+    loop_descrs[0] = given_descrs[0];
+    long in_size = ((ASCIIDTypeObject *)given_descrs[0])->size;
+    if (given_descrs[1] == NULL) {
+        PyArray_Descr *unicode_descr = PyArray_DescrNewFromType(NPY_UNICODE);
+        // numpy stores unicode as UCS4 (4 bytes wide), so bitshift
+        // by 2 to get the number of bytes needed to store the UCS4 charaters
+        unicode_descr->elsize = in_size << 2;
```
seberg, line 227: My very personal thing would be to use `* 4` and `/ 4` and trust the compiler to know that it's just a bit shift.  (not that I checked that they do :)).

asciidtype/asciidtype/src/casts.c
```diff
@@ -72,33 +76,243 @@ ascii_to_ascii(PyArrayMethod_Context *context, char *const data[],
...
+        for (int i = 0; i < copy_size; i++) {
+            if (ucs4_character_is_ascii(in) == -1) {
+                PyErr_SetString(
+                        PyExc_TypeError,
+                        "Can only store ASCII text in a ASCIIDType array.");
+                return -1;
+            }
+            // UCS4 character is ascii, so copy first byte of character
+            // into output, ignoring the rest
+            *(out + i) = *(in + i * 4);
```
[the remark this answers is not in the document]
seberg: I don't think you are trying to support unaligned access anyway anymore.  Plus, the current code is probably endianess-specific anyway, and just fails on big endian systems.

seberg (commented):
Just a few comments, I would undo the GIL grabbing thingy and simplify the unicode -> ascii loop by also casting to Py_UCS4.

PR #15, opened 2022-12-15
Fixed repr for quad dtype; added tests

peytondmurray:
Changes:

- Fixed `QuadDType` repr, so now arrays can be printed correctly
- Added tests, including CI

### DIFF
--- a/quaddtype/quaddtype/src/umath.c
+++ b/quaddtype/quaddtype/src/umath.c
@@ -65,9 +65,11 @@ quad_multiply_resolve_descriptors(PyObject *self, PyArray_DTypeMeta *dtypes[],
     Py_INCREF(given_descrs[1]);
     loop_descrs[1] = given_descrs[1];
 
-    Py_INCREF(given_descrs[1]);
-    loop_descrs[1] = given_descrs[1];
-    return NPY_SAFE_CASTING;
+    // For now, we assume that the input dtypes are all simply quads. We therefore
+    // can just reuse the input dtype for the output dtype, given by loop_descrs[2]
+    Py_INCREF(given_descrs[0]);
+    loop_descrs[2] = given_descrs[0];
+    return NPY_NO_CASTING;
 }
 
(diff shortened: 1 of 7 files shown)


quaddtype/quaddtype/src/umath.c
```diff
@@ -65,9 +65,10 @@ quad_multiply_resolve_descriptors(PyObject *self, PyArray_DTypeMeta *dtypes[],
     Py_INCREF(given_descrs[1]);
     loop_descrs[1] = given_descrs[1];
 
-    Py_INCREF(given_descrs[1]);
-    loop_descrs[1] = given_descrs[1];
-    return NPY_SAFE_CASTING;
+    // For now, assume that the input dtypes are all simply quads
```
ngoldbaum, line 68: loop_descrs[2] is the output dtype

quaddtype/tests/test_quaddtype.py
```diff
@@ -1,9 +1,22 @@
...
-    test_instantiate()
+
+def test_scalar_creation():
+    assert str(QuadScalar(3.1)) == "3.1"
+
+
+def test_create_with_explicit_dtype():
+    assert repr(
+        np.array([3.0, 3.1, 3.2], dtype=QuadDType())
+    ) == "array([3.0, 3.1, 3.2], dtype=This is a quad (128-bit float) dtype.)"
```
ngoldbaum, line 17: you could add a case that creates an array and calls `tobytes()` on it to verify that the data are actually 128 bit floats, since that's not visible from just looking at the repr

quaddtype/quaddtype/src/umath.c
peytondmurray: Thanks, I added a little more explanation of what I meant by this so it is no longer ambiguous.

quaddtype/tests/test_quaddtype.py
peytondmurray: :100: Excellent idea - see the new test I added.

PR #16, opened 2022-12-16
ASCIIDType: add implementations for add and equal ufuncs

ngoldbaum:


asciidtype/tests/test_asciidtype.py
```diff
@@ -177,3 +177,38 @@ def test_casting_fails_with_non_ascii_characters():
             match="Can only store ASCII text in a ASCIIDType array.",
         ):
             arr.astype(ASCIIDType(5))
+
+
+@pytest.mark.parametrize(
+    ("input1", "input2", "expected"),
+    [
+        (["hello "], ["world"], ["hello world"]),
+        (["a", "b", "c"], ["aa", "bbbb", "cc"], ["aaa", "bbbbb", "ccc"]),
```
seberg, line 186: This is missing a tst for variable length of the first string.  And the loop seems incorrect to me:  You will have to find how much NUL-padding the first string has and start copying the other string from there.

asciidtype/asciidtype/src/umath.c
```diff
@@ -9,4 +9,229 @@
...
+
+static NPY_CASTING
+ascii_add_resolve_descriptors(PyObject *NPY_UNUSED(self),
+                              PyArray_DTypeMeta *dtypes[],
+                              PyArray_Descr *given_descrs[],
+                              PyArray_Descr *loop_descrs[],
+                              npy_intp *NPY_UNUSED(view_offset))
+{
+    long op1_size = ((ASCIIDTypeObject *)given_descrs[0])->size;
+    long op2_size = ((ASCIIDTypeObject *)given_descrs[1])->size;
```
seberg, line 48: One of these can be `NULL` for `np.sum([...])`.  Otherwise things are fine if you return almost anything:  NumPy will complain automatically for `sum()` if (first) input and output dtypes don't match.

asciidtype/asciidtype/src/umath.c
```diff
@@ -9,4 +9,229 @@
...
+    long in2_size = ((ASCIIDTypeObject *)descrs[1])->size;
+
+    npy_intp N = dimensions[0];
+    char *in1 = data[0], *in2 = data[1], *out = data[2];
+    npy_intp in1_stride = strides[0], in2_stride = strides[1],
+             out_stride = strides[2];
+
+    while (N--) {
+        strncpy(out, in1, in1_size);
+        strncpy(out + in1_size, in2, in2_size);
```
seberg, line 31: As mentioned below this needs to be `out + in1_size - nul_padding(in1)`

seberg (commented):
LGTM, only two real comments is that reductions might fail (but that doesn't matter much at this point) and the `add()` ufunc looks incorrect for when the first array has shorter elements.

Might be nice to steal the bytes loop.  And in principle we can even do that for equality, but its also instructional to just have its own comparison loop.

(bytes/unicode do have the comparison ufuncs defined.)

asciidtype/asciidtype/src/umath.c
ngoldbaum: Good point! I hadn't thought about reductions at all up until now.

When I do `np.sum` now, I get that neither is NULL. There's [a comment](https://github.com/numpy/numpy/blob/5e0ed03b6b7a6ff05843d846eedac730f20decb7/numpy/core/src/umath/ufunc_object.c#L2795-L2800) that the first argument should be eventually though. The way it's set up right now I don't see a way for me to tell the difference from inside `resolve_descriptors` between a reduction and a regular ufunc operation. Am I missing something there?

asciidtype/tests/test_asciidtype.py
ngoldbaum: Good catch! Should be fixed now.

asciidtype/asciidtype/src/umath.c
ngoldbaum: Currently the reduction fails with the following error:

```
TypeError: the resolved dtypes are not compatible with add.reduce. Resolved (ASCIIDType(5), ASCIIDType(5), ASCIIDType(10))
```

I guess this is another place where ideally we'd be doing value-based promotion, but I guess for practicality we could return the input dtype and just truncate. However, I'd need for one of the `given_descrs` to be NULL to enable that I think,

seberg: Ah OK, so it can't actually happen right now.  The error is what I referred to and I think it is fine for now.  But of course you could calculate a "correct" length here if you knew how many elements are being summed (for strings that would be weird, but for units the count is really enough).

ngoldbaum: Yeah, agreed, getting the number of elements or array shape inside `resolve_descriptors` would be nice for reductions.

PR #20, opened 2023-01-09
fix reference counting issues in `asciidtype` and `metadatadtype`

ngoldbaum:
`common_instance` must return a new reference, but in these two dtypes I was returning a borrowed reference instead. This ended up not being a problem for `metadatadtype` because of a missing `Py_DECREF` in `get_metadata`, but was a problem for `ASCIIDType` because there isn't any indirection to get the dtype's "value" so there wasn't an opportunity to make an identical bug.

This PR fixes reference counting issues in both dtypes, adds a new test for `ASCIIDType` to check that creating an array from scalars works correctly, and fills out `common_instance` for `asciidtype` based on the `size` attribute.

### DIFF
--- a/asciidtype/asciidtype/src/dtype.c
+++ b/asciidtype/asciidtype/src/dtype.c
@@ -55,20 +55,20 @@ new_asciidtype_instance(long size)
 }
 
 /*
- * This is used to determine the correct dtype to return when operations mix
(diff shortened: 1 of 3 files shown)


ngoldbaum:
Going to self-merge, but if you're reading this and spot any issues please let me know and I'll come back to this.

PR #21, opened 2023-01-10
Initial skeleton of string dtype

peytondmurray:
This is a _very_ rough skeleton of the string dtype that stores pointers to character arrays. It's not done, but I wanted to get a PR here so that @ngoldbaum could also start using it.

Other stuff:
- Replace flake8 with ruff, it's 10-100x faster and nearly has feature parity
- `pre-commit autoupdate` because we were a bit behind on updates
- Add a config for asv benchmarks to track speedups compared to asciidtype and object arrays. I'll want to work on this more to get a comprehensive benchmarking suite up and running.


ngoldbaum:
Are you ok with calling the dtype class `StringDType` and package `stringdtype`? Just to save a little typing. I’m happy to do the search-and-replace tomorrow morning and then merge the PR if you don’t want to mess with that today.

Some TODOs:

* Add tests based on asciidtype’s
* Store UTF-8 data and cast to/from string object arrays as well as unicode arrays with non-ascii data.
* Add the remaining ufuncs that make sense to implement. I think pandas printing needs `isnan`, for one.

Probably lots of other things!

What were you planning to work ok next?

peytondmurray:
> Are you ok with calling the dtype class StringDType and package stringdtype?

:100: I don't have time now, but I'm happy to do this tomorrow when I wake up.

Mind if I take the tests first?

ngoldbaum:
OK, I’ll work on storing UTF-8 data and the casts.

peytondmurray:
Okay, it's all been replaced. If it it looks good, :ship:

PR #22, opened 2023-01-11
fixes for StringDType

ngoldbaum:
This brings over some functionality from `asciidtype`. I also include some `asciidtype` changes here to keep `stringdtype` and `asciidtype` more uniform.

* Makes `StringScalar` a `str` subclass. This makes it possible to use the functions in `np.char` to manipulate `StringDType` data
    * Adds `partition` and `rpartition` wrappers to make sure the result of those functions has uniform types.
* Adds a missing incref to `common_instance`. Also simplifies it a bit since `StringDType` isn't parametric.
* Can create a `StringDType` with an optional `size` argument, which gets ignored. This makes the API more uniform with numpy's other string dtypes. In particular, this way of creating a string dtype is now used in `np.char`, so this also allows `StringDType` to work with functions in `np.char`. Currently we just ignore the `size` but in principle we could use it as a size hint? I'm not sure if that's actually useful.

### DIFF
--- a/stringdtype/stringdtype/src/dtype.c
+++ b/stringdtype/stringdtype/src/dtype.c
@@ -21,20 +21,16 @@ new_stringdtype_instance(void)
     return new;
 }
 
-//
-// This is used to determine the correct dtype to return when operations mix
-// dtypes (I think?). For now just return the first one.
-//
+/*
+ * This is used to determine the correct dtype to return when dealing
+ * with a mix of different dtypes (for example when creating an array
+ * from a list of scalars). Since StringDType doesn't have any parameters,
+ * we can safely always return the first one.
+ */
 static StringDTypeObject *
 common_instance(StringDTypeObject *dtype1, StringDTypeObject *dtype2)
 {
-    if (!PyObject_RichCompareBool((PyObject *)dtype1, (PyObject *)dtype2,
-                                  Py_EQ)) {
-        PyErr_SetString(
-                PyExc_RuntimeError,
-                "common_instance called on unequal StringDType instances");
-        return NULL;
-    }
+    Py_INCREF(dtype1);
     return dtype1;
 }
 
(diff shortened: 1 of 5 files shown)


seberg:
Honestly, now that I see arbitrary sized strings passing length seems weird.  For most functions, the long-term thing would be a ufunc replacement.  For others, I suppose we have no good story, we could hide something away like `arr.dtype.array_funcs.char.mod`.

Anyway, don't want to derail trials here, just seems like something better is needed if we really want full support for something like `np.char`.

stringdtype/stringdtype/src/dtype.c
```diff
@@ -66,34 +62,65 @@ string_discover_descriptor_from_pyobject(PyArray_DTypeMeta *NPY_UNUSED(cls),
     return ret;
 }
 
+static PyObject *
+get_value(PyObject *scalar)
+{
+    PyObject *ret_bytes = NULL;
+    PyTypeObject *scalar_type = Py_TYPE(scalar);
+    // FIXME: handle bytes too
```
seberg, line 70: Maybe numpy bytes only?  Normal bytes shouldn't be necessary, Python 3 doesn't do it.  NumPy bytes are just weird because they serve the dual purpose of an ascii/latin1 string.

ngoldbaum:
Agreed that passing a size is weird. I'd like to come back to that, I agree that it might be better to just expose all the functionality in `np.char` as ufuncs. That just seemed like a bigger project than I wanted to take on right now.

stringdtype/stringdtype/src/dtype.c
```diff
@@ -21,20 +21,16 @@ new_stringdtype_instance(void)
...
 common_instance(StringDTypeObject *dtype1, StringDTypeObject *dtype2)
 {
-    if (!PyObject_RichCompareBool((PyObject *)dtype1, (PyObject *)dtype2,
-                                  Py_EQ)) {
-        PyErr_SetString(
-                PyExc_RuntimeError,
-                "common_instance called on unequal StringDType instances");
-        return NULL;
-    }
+    Py_INCREF(dtype1);
```
peytondmurray, line 33: :+1:

stringdtype/stringdtype/src/dtype.c
```diff
@@ -119,6 +146,15 @@ static PyType_Slot StringDType_Slots[] = {
 static PyObject *
 stringdtype_new(PyTypeObject *NPY_UNUSED(cls), PyObject *args, PyObject *kwds)
 {
+    static char *kwargs_strs[] = {"size", NULL};
+
+    long size = 0;
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "|l:ASCIIDType", kwargs_strs,
```
peytondmurray, line 153:
```suggestion
    if (!PyArg_ParseTupleAndKeywords(args, kwds, "|l:StringDType", kwargs_strs,
```

ngoldbaum: good catch!

peytondmurray (approved):
Looks good with a minor change.

PR #25, opened 2023-01-17
Unicode <-> string casts and the equal ufunc

ngoldbaum:
This adds:

* Black to the pre-commit hook. 
    * I prefer having black running so adding it to the pre-commit hook will avoid code churn introduced by my editor automatically running black on python files. Peyton said he's ok with this.
* An implementation for the equal ufunc which is used by tests. 
    * Currently it assumes all stored strings are valid null-terminated strings and simply uses `strcmp` to do the comparison.
* A `StringDType` to `np.unicode_` cast.
    * The decoder is relatively simple and doesn't do any error checking, so if the strings contain invalid UTF-8 then bad things might happen.
* A `np.unicode_` to `StringDType` cast.
    * The only error checking in the encoder is for invalid (too large) code points.
    * The output buffer is initially pessimistically `malloc`'d to be large enough to hold only 4-byte characters, I then encode code point by code point, tracking the number of bytes actually used, and call `realloc`. This approach means we build the string in one pass over the code points.
 * Tests for the new casts (and implicity for the equal ufunc, since the assertions rely on the ufunc).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -98,3 +98,12 @@ repos:
       - id: isort
         name: isort (pyi)
         types: [pyi]
+  - repo: https://github.com/psf/black
+    rev: 22.12.0
+    hooks:
+      - id: black
+        # It is recommended to specify the latest version of Python
+        # supported by your project here, or alternatively use
+        # pre-commit's default_language_version, see
+        # https://pre-commit.com/#top_level-default_language_version
+        language_version: python3.9
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -102,8 +102,3 @@ repos:
     rev: 22.12.0
     hooks:
       - id: black
-        # It is recommended to specify the latest version of Python
-        # supported by your project here, or alternatively use
-        # pre-commit's default_language_version, see
-        # https://pre-commit.com/#top_level-default_language_version
-        language_version: python3.9
--- a/stringdtype/stringdtype/src/casts.c
+++ b/stringdtype/stringdtype/src/casts.c
@@ -1,18 +1,19 @@
 #include "casts.h"
 
+#include "dtype.h"
+
 static NPY_CASTING
-string_to_string_resolve_descriptors(PyObject *NPY_UNUSED(self),
-                                     PyArray_DTypeMeta *NPY_UNUSED(dtypes[2]),
-                                     PyArray_Descr *given_descrs[2],
-                                     PyArray_Descr *loop_descrs[2],
-                                     npy_intp *NPY_UNUSED(view_offset))
+string_resolve_descriptors(PyObject *NPY_UNUSED(self),
+                           PyArray_DTypeMeta *NPY_UNUSED(dtypes[2]),
+                           PyArray_Descr *given_descrs[2],
+                           PyArray_Descr *loop_descrs[2],
+                           npy_intp *NPY_UNUSED(view_offset))
 {
     Py_INCREF(given_descrs[0]);
     loop_descrs[0] = given_descrs[0];
 
     if (given_descrs[1] == NULL) {
-        Py_INCREF(given_descrs[0]);
-        loop_descrs[1] = given_descrs[0];
+        loop_descrs[1] = (PyArray_Descr *)new_stringdtype_instance();
     }
     else {
         Py_INCREF(given_descrs[1]);
@@ -49,7 +50,7 @@ string_to_string(PyArrayMethod_Context *context, char *const data[],
 static PyArray_DTypeMeta *s2s_dtypes[2] = {NULL, NULL};
 
 static PyType_Slot s2s_slots[] = {
-        {NPY_METH_resolve_descriptors, &string_to_string_resolve_descriptors},
+        {NPY_METH_resolve_descriptors, &string_resolve_descriptors},
         {NPY_METH_strided_loop, &string_to_string},
         {NPY_METH_unaligned_strided_loop, &string_to_string},
         {0, NULL}};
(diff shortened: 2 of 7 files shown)


stringdtype/stringdtype/src/casts.c
```diff
@@ -64,12 +65,293 @@ PyArrayMethod_Spec StringToStringCastSpec = {
...
+    // 4 bytes per UCS4 character
+    npy_intp in_stride = strides[0] / 4;
+    // strides are in bytes but pointer offsets are in pointer widths, so
+    // divide by the element size (one pointer width) to get the pointer offset
+    npy_intp out_stride = strides[1] / context->descriptors[1]->elsize;
+
+    while (N--) {
+        // pessimistically allocate 4 bytes per allowed character
+        // plus one byte for the null terminator
+        char *out_buf = malloc((in_size * 4 + 1) * sizeof(char));
```
seberg, line 145: I suspect you should strip the padding first.  The side note here is that NumPy *does* actually allow included 0s, so long there is an actual character later on! 

In the place where numpy has it, there is this little thing to find the length (different context).  Maybe a `static inline` helper?
```
    /* trim trailing zeros */
    while (ucs4len > 0 && src[ucs4len - 1] == 0) {
        ucs4len--;
    }
```

That also removes the need for `realloc` which isn't pretty...

stringdtype/stringdtype/src/casts.c
```diff
@@ -64,12 +65,293 @@ PyArrayMethod_Spec StringToStringCastSpec = {
...
+
+static NPY_CASTING
+string_to_unicode_resolve_descriptors(PyObject *NPY_UNUSED(self),
+                                      PyArray_DTypeMeta *NPY_UNUSED(dtypes[2]),
+                                      PyArray_Descr *given_descrs[2],
+                                      PyArray_Descr *loop_descrs[2],
+                                      npy_intp *NPY_UNUSED(view_offset))
+{
+    Py_INCREF(given_descrs[0]);
+    loop_descrs[0] = given_descrs[0];
```
seberg, line 211: Move this to the end/after the error, so you don't have to worry about cleanup when the error path is hit.

stringdtype/stringdtype/src/casts.c
```diff
@@ -64,12 +65,293 @@ PyArrayMethod_Spec StringToStringCastSpec = {
...
+
+    if (given_descrs[1] == NULL) {
+        loop_descrs[1] = (PyArray_Descr *)new_stringdtype_instance();
```
seberg, line 79: Could fail (shouldn't really ever), also here, if you add an error check, better to move the first thing to the end.  (I know that reverse order is annoying, but it seems the easiest way for error handling)

stringdtype/stringdtype/src/casts.c
```diff
@@ -64,12 +65,293 @@ PyArrayMethod_Spec StringToStringCastSpec = {
...
+
+            // increment out_buf by the size of the character
+            out_buf += num_bytes;
+        }
+
+        // reset out_buf to the beginning of the string
+        out_buf -= out_num_bytes;
+
+        // pad string with null character
+        out_buf[out_num_bytes] = '\0';
```
seberg, line 181: For NumPy you have to pad, for your strings not ;).  I guess you don't currently have the length of the string stored at the beginning of each?  Sounds like a good addition, but also maybe something for later...

seberg (commented):
Just had a very shallow scroll-through, but I like what I saw :).  Except that realloc ;), stripping trailing 0s first is both more correct and cleaner code, I think.

stringdtype/stringdtype/src/casts.c
ngoldbaum: Yeah I was thinking today about whether we should store a struct or use some other trick to keep the length and completely avoid the possibility of buffer overruns.

ngoldbaum:
Thanks for looking over the code!

stringdtype/stringdtype/src/casts.c
seberg: I guess something like `{int length; char *buf[]}` would be the base struct.  Then you just allocate the size with `sizeof(struct) + length_of_string` and can cast/access this nicely.
I guess you could do tricks to safe storage on short strings, but dunno if that is worthwhile (basically, the first unicode character could actually be the length).

ngoldbaum: It’s a lot of code to vendor and we probably don’t need most of the features, but `sds`[https://github.com/antirez/sds#sds-internals-and-advanced-usage] might be worth looking at, in particular the way the library simply returns a `char *` but uses [a trick](https://github.com/antirez/sds#sds-internals-and-advanced-usage) to ensure there’s always a valid header before the string.

seberg: Ah, they have a free also to allow growing.  I guess storing the actual start can nice in practice.  If it is useful, vendoring seems fine to me (or submodule)?

I suppose we can invent an arbitrary amount of nifty storage schemes :), if you store pointer+length, then you have at least 10 bytes for short strings to store without any allocation ;).

.pre-commit-config.yaml
```diff
@@ -98,3 +98,12 @@ repos:
...
         types: [pyi]
+  - repo: https://github.com/psf/black
+    rev: 22.12.0
+    hooks:
+      - id: black
+        # It is recommended to specify the latest version of Python
+        # supported by your project here, or alternatively use
+        # pre-commit's default_language_version, see
+        # https://pre-commit.com/#top_level-default_language_version
+        language_version: python3.9
```
peytondmurray, line 109: Hmm, is there any way to make this more flexible? I'm developing with 3.11...

ngoldbaum: I can try this with the default language version, I don’t think it actually changes the formatting. I just copy/pasted this snippet from the black docs.

stringdtype/stringdtype/src/casts.c
peytondmurray: I also like the idea of having a struct of `{int length; char *buf[]}`, I think it would make the logic here nicer. If vendoring seems best to everyone else here, I'm all for it :)

stringdtype/stringdtype/src/casts.c
ngoldbaum: Thanks for the snippet, it is useful! I don't think doing this completely eliminates the need for the `realloc`, but I can avoid using it if I first strip trailing zeros in the UCS4 sequence then do an initial pass over the remaining codepoints to calculate the number of bytes needed to store the UTF8 string.

Did you have another trick in mind when you wrote this comment that would let me do the conversion with only one pass over the codepoints and one call to `malloc`?

ngoldbaum:
I'm going to punt on changing the string storage to a followup. I think `sds` might be a little too much for what we need, since I don't think we need dynamic strings at all, so I'm going to make something following it but a little simpler (simple static strings, `sss`? :wink:).

If there's a way to do the conversion from codepoints to UTF-8 with a single pass over the codepoints and only one `malloc` (no `realloc`) I'd love to hear about it, but I don't think that's possible, so the last commit adds a first pass over the codepoints to calculate the size of the output buffer. I also moved the error checking into that function.

stringdtype/stringdtype/src/casts.c
seberg: Ah, you are right of course. you have to do two passes two find the byte-length with unicode encoding.  Not sure what the best/typical way to do this is, I bet there are nifty tricks.  Might want to check if `PyUnicode` does anything fun when it needs to convert to utf-8?

peytondmurray (approved):
Looks good, thank you for this! :rocket:

PR #27, opened 2023-01-20
Add ufunc promoter for equal

ngoldbaum:
This adds a promoter implementation for the equal ufunc. This makes it so that if you pass a fixed-width unicode string dtype and a `StringDType` array to `np.equal`, numpy will automatically cast the `Unicode` to `StringDType`.

@seberg is there a way to do this without adding the promoter twice? If not, would it make sense for `PyUFunc_AddPromoter` to add two promoters, one with the input arguments switched? I tried only doing it once and the promoter only kicked in for when `StringDType` was the first parameter, so I think the order does matter currently.

`default_ufunc_promoter` is copy/pasted from numpy. Should I be doing something a little more precise?

### DIFF
--- a/stringdtype/stringdtype/src/umath.c
+++ b/stringdtype/stringdtype/src/umath.c
@@ -135,7 +135,7 @@ init_equal_ufunc(PyObject *numpy)
     }
 
     /*
-     * Initialize spec for equality
+     *  Initialize spec for equality
      */
     PyArray_DTypeMeta *eq_dtypes[3] = {&StringDType, &StringDType,
                                        &PyArray_BoolDType};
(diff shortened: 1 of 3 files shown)


ngoldbaum:
> If not, would it make sense for PyUFunc_AddPromoter to add two promoters, one with the input arguments switched?

I guess this would only work for ufuncs where the order of the inputs don't matter, so `PyUFunc_AddPromoter` can't always just do it.

seberg:
> I guess this would only work for ufuncs where the order of the inputs don't matter, so PyUFunc_AddPromoter can't always just do it.

Exactly, multiple-dispatching is complicated :/.  There could be a point in signalling "I don't care about order", but...  In the end for anything but simple binary cases like this one, the only good solution right now I have is the abstract dtype.
(Which is also tricky, cause you can't say "one of them must be mystring", so your promoter has to do the right thing for NumPy strings also in principle...)

stringdtype/stringdtype/src/umath.c
```diff
@@ -73,36 +137,74 @@ init_equal_ufunc(PyObject *numpy)
...
-        free(EqualSpec);
+    PyObject *DTypes1 = PyTuple_Pack(3, &StringDType, &PyArray_UnicodeDType,
+                                     &PyArrayDescr_Type);
+    if (DTypes1 == 0) {
+        Py_DECREF(promoter_capsule1);
+        return -1;
+    }
+
+    if (PyUFunc_AddPromoter(equal, DTypes1, promoter_capsule1) < 0) {
+        Py_DECREF(promoter_capsule1);
```
seberg, line 180: You can use the same promoter twice, nothing is prevnting you from that.  Otherwise LGTM (and happy either way).

PR #28, opened 2023-01-23
update to new DType API version

ngoldbaum:
https://github.com/numpy/numpy/pull/20970 bumped the DType API version, so let's bump the API version here as well to fix the tests. In principle some of these dtypes could use the new API slot, but going to punt on that for now unless there's new test breakage.


ngoldbaum:
Going to self-merge quickly to unbreak the tests. If anyone has any issues with that let me know and we can hash things out.

seberg:
Thanks!

PR #30, opened 2023-01-24
store string data in a struct along with length

ngoldbaum:
This makes the dtype store the string data in a struct:

```C
typedef struct ss {
    size_t len;
    char buf[];
} ss;
```

So far this is just a refactoring but the hope is that storing the length will avoid O(N) `strlen` calls and make it easier to write code that doesn't accidentally introduce buffer overflows.

Bikeshedding on the naming and the API are very welcome.

### DIFF
--- a/stringdtype/stringdtype/src/dtype.c
+++ b/stringdtype/stringdtype/src/dtype.c
@@ -1,6 +1,7 @@
 #include "dtype.h"
 
 #include "casts.h"
+#include "static_string.h"
 
 PyTypeObject *StringScalar_Type = NULL;
 
@@ -15,8 +16,8 @@ new_stringdtype_instance(void)
     if (new == NULL) {
         return NULL;
     }
-    new->base.elsize = sizeof(char *);
-    new->base.alignment = _Alignof(char *);
+    new->base.elsize = sizeof(ss *);
+    new->base.alignment = _Alignof(ss *);
 
     return new;
 }
@@ -113,8 +114,7 @@ stringdtype_setitem(StringDTypeObject *descr, PyObject *obj, char **dataptr)
         return -1;
     }
 
-    *dataptr = malloc(sizeof(char) * length + 1);
-    strncpy(*dataptr, val, length + 1);
+    *dataptr = (char *)ssnewlen(val, length);
     Py_DECREF(val_obj);
     return 0;
 }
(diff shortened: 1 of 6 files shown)


stringdtype/stringdtype/src/dtype.c
```diff
@@ -113,16 +114,20 @@ stringdtype_setitem(StringDTypeObject *descr, PyObject *obj, char **dataptr)
...
     }
 
-    *dataptr = malloc(sizeof(char) * length + 1);
-    strncpy(*dataptr, val, length + 1);
+    ss *str_val = ssnewlen(val, length);
+    if (str_val == NULL) {
+        PyErr_SetString(PyExc_MemoryError, "ssnewlen failed");
+        return -1;
+    }
+    *dataptr = (char *)str_val;
```
ngoldbaum, line 122: Is this a memory leak? How do I tell the difference between when I'm filling empty invalid data and when I'm overriding an existing string that I need to free first.

seberg: Set the `NEEDS_INIT` flag and check for `NULL` (currently, I only guarantee NULLing of data as initialization, I somewhat hope that is sufficient).

seberg:
Storing the length explicitly also means you can represent strings like `"a\0"` correctly (NumPy strings cannot).

I don't feel like bike-shedding storage.  There are probably many options.  If there is a chance of wanting to play with schemes that store length in the array itself then using accessor `static inline` functions or macros might make sense for getting length and chars/buf.

In theory `int` is enough for length storage for the forseeable future, but I guess it doesn't matter.

stringdtype/stringdtype/src/umath.c
```diff
@@ -29,7 +30,18 @@ string_equal_strided_loop(PyArrayMethod_Context *context, char *const data[],
...
+        size_t len1 = (*in1)->len;
+        size_t len2 = (*in2)->len;
+        size_t maxlen;
+
+        if (len1 > len2) {
+            maxlen = len1;
+        }
+        else {
+            maxlen = len2;
+        }
```
peytondmurray, line 42: Can you short-circuit the `strncmp` if the `ss` lengths are different?

peytondmurray (approved):
Optional small change :rocket: thanks for this!

PR #32, opened 2023-01-27
add asan to stringdtype build

ngoldbaum:
This builds stringdtype with asan support. I also pass `-s` to pytest so pytest doesn't capture the output from asan when a test fails.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,9 +67,14 @@ jobs:
       - name: Install stringdtype
         working-directory: stringdtype
         run: |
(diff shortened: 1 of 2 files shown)


ngoldbaum:
Both python itself and `stringdtype` leak memory, so I'm turning off `LeakSanitizer`. In the future when we fix the memory leaks in `stringdtype` we can come back to this and add suppressions for the memory leaks in python.

ngoldbaum:
looks like it's working! https://github.com/numpy/numpy-user-dtypes/actions/runs/4027429067/jobs/6923148711

PR #33, opened 2023-02-01
Add support for pickling ASCIIDtype and StringDType instances

ngoldbaum:
This adds `__setstate__` and `__reduce__` implementations for `StringDType` and `ASCIIDType`.

Currently pickling `StringDType` arrays is broken. It will work within a single python session so long as the original array still exists, but won't work between python sessions. To get it to work we'll need a hook in the DType API to serialize the string data.


asciidtype/asciidtype/src/dtype.c
```diff
@@ -226,6 +226,82 @@ static PyMemberDef ASCIIDType_members[] = {
...
+
+    PyTuple_SET_ITEM(ret, 0, obj);
+
+    PyTuple_SET_ITEM(ret, 1, Py_BuildValue("(l)", self->size));
+
+    state = PyTuple_New(1);
+
+    PyTuple_SET_ITEM(state, 0, PyLong_FromLong(PICKLE_VERSION));
+
+    PyTuple_SET_ITEM(ret, 2, state);
```
seberg, line 262: I think:
```
Py_BuildValue("O(l)", Py_TYPE(self) self->size,self->size)
```
is good enough? presumably we can consider `ASCIIDType(length)` to be stable API?  And you made the type pickable using the copyreg, so no need for the reconstruct function?  Plus, if we _have_ a reconstruction function, then we certainly can assume stable API?

asciidtype/asciidtype/src/dtype.c
```diff
@@ -226,6 +226,82 @@ static PyMemberDef ASCIIDType_members[] = {
...
+    if (version != PICKLE_VERSION) {
+        PyErr_Format(PyExc_ValueError,
+                     "Pickle version mismatch. Got version %d but expected "
+                     "version %d.",
+                     version, PICKLE_VERSION);
+        return NULL;
+    }
+
+    Py_RETURN_NONE;
+}
```
seberg, line 287: Not convinced we need it, and I would try to avoid it :).

asciidtype/asciidtype/src/dtype.c
seberg: Ah, you didn't do the copyreg for the ASCIIDtype, I think.  But maybe that is nicer?  For the StringDType, this seems true even more so?  We can assume `StringDType()` will always give the same thing, so no need for `__setstate__` logic?

seberg (commented):
I do wonder if you can't slash this all down and just rely on stable API for the class, otherwise should be fine (didn't scan for refcount leaks)

asciidtype/asciidtype/src/dtype.c
ngoldbaum: I'm not using `copyreg` at all in either dtype. It is a good point though that we don't actually need to save the dtype itself at all though, thanks for the hint.

seberg: Ah, I misread... somehow I thought one of the reconstructs was to make the class work.  So I guess you need the reconstruct helper until we fix the class pickling.

ngoldbaum: That's exactly right. I originally had it as you suggested, but that didn't work because `ASCIIDType` is an instance of `DTypeMeta`, so python checks the type, sees it's in the `copyreg`, and dutifully calls the pickler numpy registers for the type.

asciidtype/asciidtype/src/dtype.c
ngoldbaum: Are you saying we don't need to save a pickle version? I thought that was generally useful for forward compatibility, just in case we ever want to change what gets pickled in the future.

seberg: I am saying there is no point in `__setstate__`.

ngoldbaum: But then how do I check the pickle version when I load the pickle?

seberg: Yeah, if you want a version, then it is not possible.  But, you don't need a version if you assume your API is stable, and if you have a loader function, you can keep the old one around also?
Anyway, not a strong opinion, the `ndarray.__setstate__` in NumPy is just so messy, that it probably makes me want to not use it if possible :).

peytondmurray:
Would also be nice to have a test, something that pickles/unpickles would be good:

```python
def test_pickle(string_list):
    sarr = np.array(string_list, dtype=StringDType())

    pickled = pickle.dumps(sarr)
    obj = pickle.loads(pickled)

    np.testing.assert_array_equal(
        sarr, obj
    )
```

ngoldbaum:
Oops, thought the nightly wheels had gotten updated but it looks like I'll still need to wait a bit before I can merge this.

seberg:
Personally would still avoid setstate (since I feel `StringDType()` and `ASCIDType(3)` can both be considered stable API), but fair.
I just manually retrigger the wheel build, it runs only twice a week by default.

ngoldbaum:
I agree they probably won’t change, I’d just prefer to have the flexibility just in case.

PR #37, opened 2023-02-09
Add string comparison support so that x.sort() works

peytondmurray:
This PR builds on https://github.com/numpy/numpy/pull/23173 to provide a [`PyArray_CompareFunc`](https://numpy.org/devdocs/reference/c-api/types-and-structures.html#c.PyArray_ArrFuncs.compare) so that `x.sort()` works.

I also updated the README with a bit of information about needing the nightly numpy wheel for support, added a note about `--no-build-isolation`, and removed a bit about `meson-python` not supporting editable builds (it does now!).

This PR remains in draft until https://github.com/numpy/numpy/pull/23173 is merged.

### DIFF
--- a/stringdtype/stringdtype/src/dtype.c
+++ b/stringdtype/stringdtype/src/dtype.c
@@ -155,6 +155,16 @@ stringdtype_getitem(StringDTypeObject *descr, char **dataptr)
     return res;
 }
 
+// Implementation of PyArray_CompareFunc.
+// Compares unicode strings by their code points.
+int
+compare_strings(char **a, char **b, PyArrayObject *NPY_UNUSED(arr))
+{
+    ss *ss_a = (ss *)*a;
+    ss *ss_b = (ss *)*b;
+    return strcmp(ss_a->buf, ss_b->buf);
+}
+
 static StringDTypeObject *
 stringdtype_ensure_canonical(StringDTypeObject *self)
 {
@@ -170,6 +180,7 @@ static PyType_Slot StringDType_Slots[] = {
         {NPY_DT_setitem, &stringdtype_setitem},
         {NPY_DT_getitem, &stringdtype_getitem},
         {NPY_DT_ensure_canonical, &stringdtype_ensure_canonical},
+        {NPY_DT_PyArray_ArrFuncs_compare, &compare_strings},
         {0, NULL}};
 
 static PyObject *
--- a/stringdtype/tests/test_stringdtype.py
+++ b/stringdtype/tests/test_stringdtype.py
@@ -161,3 +161,23 @@ def test_pickle(string_list):
     assert res[1] == dtype
 
     os.remove(f.name)
+
+
+@pytest.mark.parametrize(
+    "strings",
+    [
+        ["left", "right", "leftovers", "righty", "up", "down"],
+        ["🤣🤣", "🤣", "📵", "😰"],
+        ["🚜", "🙃", "😾"],
+        ["😹", "🚠", "🚌"],
+        ["A¢☃€ 😊", " A☃€¢😊", "☃€😊 A¢", "😊☃A¢ €"],
+    ],
+)
+def test_sort(strings):
+    """Test that sorting matches python's internal sorting."""
+    arr = np.array(strings, dtype=StringDType())
+    arr_sorted = np.array(sorted(strings), dtype=StringDType())
+
+    np.random.default_rng().shuffle(arr)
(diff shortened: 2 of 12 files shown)


stringdtype/stringdtype/src/dtype.c
```diff
@@ -277,6 +278,15 @@ static PyMethodDef StringDType_methods[] = {
         {NULL},
 };
 
+// Implementation of PyArray_CompareFunc.
+int
+compare_strings(char **a, char **b, PyArrayObject *arr)
+{
+    ss *ss_a = (ss *)*a;
+    ss *ss_b = (ss *)*b;
+    return strcmp(ss_a->buf, ss_b->buf);
```
ngoldbaum, line 287: Rather than using `strcmp`, first check if the lengths are identical and then if they are, use the length to do `strncmp`.

stringdtype/stringdtype/src/dtype.h
```diff
@@ -26,6 +26,9 @@ new_stringdtype_instance(void);
 int
 init_string_dtype(void);
 
+int
+compare_strings(char **a, char **b, PyArrayObject *arr);
+
```
ngoldbaum, line 31: I don't think this needs to be in the header, since its only usage is inside `dtype.c`.

ngoldbaum: You could add a forward declaration inside that file, or just move its declaration above its usage.

ngoldbaum:
It occurs to me that it may not be correct to rely on `strcmp` to reliably sort utf-8 encoded data in a sensible way. It may be a good idea to look up edge cases related to sorting unicode strings to see how this approach handles them.

stringdtype/stringdtype/src/dtype.c
ngoldbaum: Oops, discard that suggestion, it only makes sense for equality comparison, not for generic comparisons.

mattip:
Unicode string comparison [is tricky](https://www.unicode.org/faq/normalization.html). There are lots of edge cases.

peytondmurray:
Yep, you're both right - `strcmp` will work for the ASCII dataset but a more robust approach seems important here. I'll spend some time looking into this.

peytondmurray:
After some discussion I've made the choice to sort strings by their UCS4 code points. This is the approach that python takes by default, and if you want the [more advanced sorting methods required for unicode comparison, you can use the `unicodedata` package.](https://docs.python.org/3/howto/unicode.html#comparing-strings)

I still think it would be cool to have vectorized functions for normalization and sorting, but we can introduce those later, since you can't provide the options you would need to do that for `arr.sort()` here anyway.

stringdtype/tests/test_stringdtype.py
```diff
@@ -14,6 +14,11 @@ def string_list():
     return ["abc", "def", "ghi"]
 
 
+@pytest.fixture
+def string_list_similar():
+    return ["left", "right", "leftovers", "righty", "up" "down"]
```
ngoldbaum, line 19: Can we get some tests with non-ascii characters?

I also suspect that you'll find when you do this that you could replace your codepoint sorting implementation with just `strcmp` and it will have identical results.

To see why this has to be the case, take a look at [the table](https://en.wikipedia.org/wiki/UTF-8#Encoding) in the wikipedia article on utf-8 encoding. The largest ascii character is `0b01111111`, but the first byte of a multicharacter string will always be at least `0b11000000`. The first bytes are also chosen so that a three-byte character will always sort after a two-byte character, ditto for four-byte strings and three-byte characters. The bits of the codepoint itself are always left in order, although the bits might be split up by continuation bits, but since the continuation bits are always the same, that doesn't affect ordering.

Unfortunately I'm having a heck of a time finding an "official" citation for this fact or even a full explanation for why it works beyond the explanation I just wrote, but I'm pretty sure it's true.

ngoldbaum:
Besides the inline comment, can you also increment the API version in the other dtypes? You'll need to do that anyway before this is merged to get the tests to pass.

stringdtype/tests/test_stringdtype.py
peytondmurray: You're right - it looks like code points and their corresponding utf8 representations both increase monotonically. I'll swap back to `strcmp` for simplicity, speed, and compactness.

peytondmurray:
Thanks for the feedback - I've updated the API versions everywhere.

PR #39, opened 2023-02-17
Refactor metadatadtype to use ufunc wrapping

ngoldbaum:
Rather than implementing ufunc loops manually, it's much more straightforward for metadatadtype to simply wrap float64 ufunc loops. This PR refactors the ufunc implementation in metadatadtype to use the `PyUFunc_AddWrappingLoop` API to do that.

Most of this is adapted from the scaled float dtype in numpy.

### DIFF
--- a/metadatadtype/metadatadtype/src/umath.c
+++ b/metadatadtype/metadatadtype/src/umath.c
@@ -12,96 +12,78 @@
 #include "umath.h"
 
 static int
-metadata_multiply_strided_loop(PyArrayMethod_Context *context,
-                               char *const data[], npy_intp const dimensions[],
-                               npy_intp const strides[], NpyAuxData *auxdata)
+translate_given_descrs_to_double(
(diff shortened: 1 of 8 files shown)


metadatadtype/metadatadtype/src/umath.c
```diff
@@ -12,96 +12,120 @@
...
+        return NULL;
     }
-    PyObject *multiply = PyObject_GetAttrString(numpy, "multiply");
-    Py_DECREF(numpy);
-    if (multiply == NULL) {
-        return -1;
+    PyObject *ufunc = PyObject_GetAttrString(mod, ufunc_name);
+    Py_DECREF(mod);
+    if (ufunc == NULL) {
+        return NULL;
```
peytondmurray, line 88:
```suggestion
```

peytondmurray (approved):
One small change, otherwise looks good!

PR #40, opened 2023-02-22
Add clear loop implementation

ngoldbaum:
Note that this can't be merged until https://github.com/numpy/numpy/pull/23262 is merged and there's a wheel available to install.

There are a couple other minor cleanups as well, but mostly this implements custom `get_clear_loop` and `clear_loop` implementations, allowing us to call `free` on the embedded references stored in the numpy array before the array is deallocated.

I haven't checked, but this will hopefully fix all memory leaks, although there may be other places we need to call `free` as well. In any case, it makes it much easier to check for those places!

### DIFF
--- a/stringdtype/stringdtype/src/dtype.c
+++ b/stringdtype/stringdtype/src/dtype.c
@@ -21,6 +21,7 @@ new_stringdtype_instance(void)
     new->base.flags |= NPY_NEEDS_INIT;
     new->base.flags |= NPY_LIST_PICKLE;
     new->base.flags |= NPY_ITEM_REFCOUNT;
+    new->base.flags |= NPY_NEEDS_PYAPI;
 
     return new;
 }
--- a/stringdtype/stringdtype/src/dtype.c
+++ b/stringdtype/stringdtype/src/dtype.c
@@ -183,6 +183,7 @@ stringdtype_clear_loop(void *NPY_UNUSED(traverse_context),
     while (size--) {
         if (data != NULL) {
             free(*(ss **)data);
+            *(ss **)data = NULL;
         }
         data += stride;
     }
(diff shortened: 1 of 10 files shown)


stringdtype/stringdtype/src/dtype.c
```diff
@@ -162,6 +164,36 @@ stringdtype_ensure_canonical(StringDTypeObject *self)
...
+static int
+stringdtype_get_clear_loop(void *NPY_UNUSED(traverse_context),
+                           PyArray_Descr *NPY_UNUSED(descr),
+                           int NPY_UNUSED(aligned),
+                           npy_intp NPY_UNUSED(fixed_stride),
+                           traverse_loop_function **out_loop,
+                           NpyAuxData **NPY_UNUSED(out_auxdata),
+                           NPY_ARRAYMETHOD_FLAGS *flags)
+{
+    *flags = NPY_METH_REQUIRES_PYAPI | NPY_METH_NO_FLOATINGPOINT_ERRORS;
```
seberg, line 192:
```suggestion
    *flags = NPY_METH_NO_FLOATINGPOINT_ERRORS;
```
You don't require the python API actually if you are u sing `free()` to free/allocate.

NumPy currently actually ignores the float flag also, because how will clearing cause that?

seberg (approved):
Looks good, besides those two comments.

stringdtype/stringdtype/src/dtype.c
ngoldbaum: Oh, I thought `NPY_METH_REQUIRES_PYAPI` was required for the clear loop, I'll try without.

seberg: For all numpy ones, yes :).

ngoldbaum:
To avoid a seg fault in the sorting routine I needed to add the `NPY_NEEDS_PYAPI` flag. There's no reason why we can't release the GIL, but to get this to work we'd need to add support in numpy for dtypes that hold references that aren't python objects (e.g. make it so `NPY_ITEM_IS_POINTER` can be set without `NPY_ITEM_REFCOUNT` while still getting data NULLing at init and the ability to call the clear function.

ngoldbaum:
This PR needs to be merged to fix the tests, but I'll hold off on self-merging just in case Sebastian or Peyton wants to weigh in on the last commit I just pushed.

seberg:
Meh... the numpy sorting code is being stupid there, but I guess you can workaround it here for the time being.

ngoldbaum:
Opened https://github.com/numpy/numpy/pull/23292 and reverted the last commit.

seberg:
I retriggered the NumPy "nightlies", so you can hopefully put this in without tests failing (unless it depended on that NULL one).

PR #41, opened 2023-02-24
Handle possible NULL pointers in the array buffer

ngoldbaum:
The discussion around https://github.com/numpy/numpy/pull/23262 made me realize that there can be NULL pointers in the array buffer after `np.zeros` and `np.empty`. This means that we need to check for that case whenever we access an existing string array buffer in the casts and ufuncs.

To support this I added a new helper `empty_if_null`, which returns a new empty static string. I couldn't figure out how to do this in a way that avoids heap allocations because the `ss` struct uses a flexible array member, so I also need to free the resulting empty string later.

It would be possible to avoid this if there was a way for a dtype to supply an initial value for `np.empty` and `np.zeros`.

Alternative suggestions for how to handle this are very welcome.

### DIFF
--- a/stringdtype/stringdtype/src/casts.c
+++ b/stringdtype/stringdtype/src/casts.c
@@ -53,11 +53,17 @@ string_to_string(PyArrayMethod_Context *context, char *const data[],
     npy_intp out_stride = strides[1] / context->descriptors[1]->elsize;
 
     while (N--) {
-        out[0] = ssdup(in[0]);
+        ss *s = empty_if_null(in);
+        out[0] = ssdup(s);
         if (out[0] == NULL) {
             gil_error(PyExc_MemoryError, "ssdup failed");
             return -1;
         }
+
+        if (*in == NULL) {
+            free(s);
+        }
+
         in += in_stride;
         out += out_stride;
     }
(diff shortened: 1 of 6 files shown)


stringdtype/stringdtype/src/casts.c
```diff
@@ -53,11 +53,17 @@ string_to_string(PyArrayMethod_Context *context, char *const data[],
...
+        ss *s = empty_if_null(in);
+        out[0] = ssdup(s);
         if (out[0] == NULL) {
             gil_error(PyExc_MemoryError, "ssdup failed");
             return -1;
         }
+
+        if (*in == NULL) {
+            free(s);
+        }
```
seberg, line 65: So, if we definitely don't do reference counting (each array that is not a view owns a copy of the string), then I do still think it makes sense to move the length information into the array data itself.

On this case, how about a pattern off:
```
    const ss *s = get_string(in);
```
And then return a statically allocated empty `ss` (or a once allocated global).  That way you don't need the `free()` call.  Freeing is only needed on store, and there you can do it unconditionally.  The only thing is you must not free what `get_string` returns.  I hope the const will protect a bit and it should at least be a nice hard crash that is easy to track down.

ngoldbaum:
Closing this in favor of a new PR implementing Sebastian's suggestion.

PR #42, opened 2023-02-28
Store contents of static string struct in the array buffer

ngoldbaum:
This is a followup for #41, since I had to make a number of extensive changes relative to what that PR originally proposed, I issued a new PR instead to hopefully make this clearer.

Instead of using flexible-length arrays, the `ss` struct now looks like this:

```
typedef struct ss {
    size_t len;
    char *buf;
} ss;
```

Rather than store a pointer to an `ss` struct in the array buffer, this refactors things so we instead store the contents of the `ss` struct in the array buffer. This reduces the amount of pointer indirection we need to do to get access to the `ss` struct data and also makes it much easier to deal with empty strings, since with this new definition `((ss*)NULL)->len` is zero by construction.

Instead of `malloc`ing storage for the entire struct on the heap, we now only `malloc` space for the string buffer, and the caller of functions that allocate memory are responsible for handling memory for the `ss` struct.

To handle this change, I had to rewrite the static_string API to accept pointers to `ss` instances instead of returning them. I also added a new `load_string` helper that returns either a read-only view onto an `ss` struct, or if the struct is `NULL`, returns a view onto a statically allocated empty struct.

### DIFF
--- a/stringdtype/stringdtype/src/dtype.c
+++ b/stringdtype/stringdtype/src/dtype.c
@@ -125,9 +125,11 @@ stringdtype_setitem(StringDTypeObject *NPY_UNUSED(descr), PyObject *obj,
 
     // copies contents of val into item_val->buf
     int res = ssnewlen(val, length, (ss *)dataptr);
+
     // val_obj must stay alive until here to ensure *val* doesn't get
     // deallocated
     Py_DECREF(val_obj);
+
     if (res == -1) {
         PyErr_NoMemory();
         return -1;
@@ -137,7 +139,6 @@ stringdtype_setitem(StringDTypeObject *NPY_UNUSED(descr), PyObject *obj,
         assert(0);
     }
 
-    Py_DECREF(val_obj);
     return 0;
 }
 
--- a/stringdtype/stringdtype/src/dtype.c
+++ b/stringdtype/stringdtype/src/dtype.c
@@ -124,13 +124,19 @@ stringdtype_setitem(StringDTypeObject *NPY_UNUSED(descr), PyObject *obj,
     ssfree((ss *)dataptr);
 
     // copies contents of val into item_val->buf
-    if (ssnewlen(val, length, (ss *)dataptr) == -1) {
-        PyErr_SetString(PyExc_MemoryError, "ssnewlen failed");
-        return -1;
-    }
-
+    int res = ssnewlen(val, length, (ss *)dataptr);
     // val_obj must stay alive until here to ensure *val* doesn't get
     // deallocated
+    Py_DECREF(val_obj);
+    if (res == -1) {
+        PyErr_NoMemory();
+        return -1;
+    }
+    else if (res == -2) {
+        // this should never happen
+        assert(0);
+    }
+
     Py_DECREF(val_obj);
     return 0;
 }
@@ -155,10 +161,6 @@ stringdtype_getitem(StringDTypeObject *NPY_UNUSED(descr), char **dataptr)
 
     PyObject *res = PyObject_CallFunctionObjArgs((PyObject *)StringScalar_Type,
                                                  val_obj, NULL);
-
-    if (res == NULL) {
-        return NULL;
-    }
 
     Py_DECREF(val_obj);
 
--- a/stringdtype/stringdtype/src/dtype.c
+++ b/stringdtype/stringdtype/src/dtype.c
@@ -16,8 +16,8 @@ new_stringdtype_instance(void)
     if (new == NULL) {
         return NULL;
     }
-    new->base.elsize = sizeof(ss *);
-    new->base.alignment = _Alignof(ss *);
+    new->base.elsize = sizeof(ss);
+    new->base.alignment = _Alignof(ss);
     new->base.flags |= NPY_NEEDS_INIT;
     new->base.flags |= NPY_LIST_PICKLE;
     new->base.flags |= NPY_ITEM_REFCOUNT;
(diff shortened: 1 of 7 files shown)


stringdtype/stringdtype/src/dtype.c
```diff
@@ -119,26 +119,35 @@ stringdtype_setitem(StringDTypeObject *NPY_UNUSED(descr), PyObject *obj,
...
-    ss *str_val = ssnewlen(val, length);
-    if (str_val == NULL) {
+    // free if dataptr holds preexisting string data,
+    // ssfree does a NULL check
+    ssfree((ss *)dataptr);
+
+    // copies contents of val into item_val->buf
+    if (ssnewlen(val, length, (ss *)dataptr) == -1) {
         PyErr_SetString(PyExc_MemoryError, "ssnewlen failed");
         return -1;
```
seberg, line 129:
```suggestion
        Py_DECREF(val_obj);
        return -1;
```

seberg: or use the pattern:
```
res = ssnewlen(val, length, (ss *)dataptr)
Py_DECREF(val);
if (res < 0) {
    PyErr_NoMemory();
    return -1;
}
return 0;
```

stringdtype/stringdtype/src/dtype.c
```diff
@@ -119,26 +119,35 @@ stringdtype_setitem(StringDTypeObject *NPY_UNUSED(descr), PyObject *obj,
...
+        data = "\0";
+    }
+    else {
+        data = ((ss *)dataptr)->buf;
+    }
+
+    PyObject *val_obj = PyUnicode_FromString(data);
 
     if (val_obj == NULL) {
         return NULL;
```
seberg, line 153: A few lines down also, `Py_DECREF(val_obj);` should go directly after the call.

(Then you can also just `return res`).

stringdtype/stringdtype/src/static_string.c
```diff
@@ -1,41 +1,76 @@
...
+
+void
+load_string(char *data, ss **out)
+{
+    if (data == NULL) {
+        *out = &EMPTY;
+    }
+    else {
+        *out = (ss *)data;
+    }
```
seberg, line 75: I don't think you should need this helper now?  data should never be NULL here? `data->buf` may be, but that is OK.

OTOH, I don't mind having the helper, it might be neat to experiment a bit more easily also.

seberg (approved):
Didn't have a thorough look through, but I like this, thanks.

stringdtype/stringdtype/src/static_string.c
```diff
@@ -1,41 +1,76 @@
...
-    ss *ret = (ss *)malloc(sizeof(ss) + sizeof(char) * (len + 1));
-    ret->len = len;
-    return ret;
+    if (out->len != 0 || out->buf != NULL) {
+        return -1;
+    }
+
+    char *buf = (char *)malloc(sizeof(char) * (num_bytes + 1));
+
+    if (buf == NULL) {
```
peytondmurray, line 55: In this case is it appropriate to set a python exception? This happens when memory can't be allocated.

stringdtype/stringdtype/src/static_string.h
```diff
@@ -6,21 +6,40 @@
...
+// *out*, checks that *out->buf* is NULL and *out->len* is zero and errors
+// otherwise. Also errors if malloc fails.
+int
+ssdup(ss *in, ss *out);
+
+// Allocates a new string buffer for *out* with enough capacity to store
+// *num_bytes* of text. The actual allocation is num_bytes + 1 bytes, to
+// account for the null terminator. Does not do any initialization, the caller
+// must initialize and null-terminate the string buffer. Errors if *out*
+// already contains data or if malloc fails.
```
peytondmurray, line 32: :rocket: This is super valuable documentation. Thank you!

peytondmurray (approved):
Reducing the number of stars everywhere makes this so much more readable, thanks for this! :star:

stringdtype/stringdtype/src/static_string.c
ngoldbaum: I'd prefer to leave it to the responsibility of the caller to check the error code. That way this `static_string` library doesn't need to depend on `python.h` or worry about the GIL.

That said, I'm going to make it so you get back `-2` if `out` is not empty and `-1` if malloc fails.

stringdtype/stringdtype/src/static_string.c
ngoldbaum: Yeah, my worry was what if code uses `data->buf`, even if `data->len` is zero. In practice that likely won't happen and if it does we should get a seg fault so it'll be really obvious.

stringdtype/stringdtype/src/static_string.c
peytondmurray: Sounds good.

stringdtype/stringdtype/src/static_string.c
seberg: Ah, I guess since you are currently using/guaranteeing `\0` terminated strings it kind-of makes sense.

seberg: But, you are not looking for `data == NULL` in that case, you would be looking for `data->len == 0` or `data->buf == NULL`.

ngoldbaum: Argh, good call!

PR #45, opened 2023-03-03
Fix memory leak in _memory_usage

ngoldbaum:
Also removes an incorrect comment I found while working on https://github.com/numpy/numpy/pull/23318

### DIFF
--- a/stringdtype/stringdtype/src/dtype.c
+++ b/stringdtype/stringdtype/src/dtype.c
@@ -246,8 +246,6 @@ stringdtype_new(PyTypeObject *NPY_UNUSED(cls), PyObject *args, PyObject *kwds)
 static void
 stringdtype_dealloc(StringDTypeObject *self)
 {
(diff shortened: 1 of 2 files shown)


peytondmurray:
Thanks for finding this, this seemed like a difficult one. During your debugging, did you find a good way that we can test for something like this?

ngoldbaum:
Yeah, we can add a `pytest-leaks` run or just make all the tests run under `pytest-leaks` (which willake it slower). That would require a decent amount of messing with the github actions setup since it needs a debug python build and I wanted to look at other stuff today.

seberg:
`pytest-valgrind` would also find true memory leaks (not refcounting problems, pytest-leaks is more sensitive).  Similar speed issue, although easier to set up and less fragile.

(anyway, just throwing it out, overall, testing occasionally with sanitizers is likely even better these days.)

PR #46, opened 2023-03-03
Add a cast from string to bool

ngoldbaum:
First, this refactors the string casting setup a bit to make it more general and have a little less boilerplate. Second, it adds a cast from string to bool, which notably makes `np.any` and `np.all` work correctly.

### DIFF
--- a/stringdtype/stringdtype/src/casts.c
+++ b/stringdtype/stringdtype/src/casts.c
@@ -77,8 +77,6 @@ static PyType_Slot s2s_slots[] = {
 
 static char *s2s_name = "cast_StringDType_to_StringDType";
 
-static PyArray_DTypeMeta *s2s_dtypes[] = {NULL, NULL};
-
 // unicode to string
 
 static NPY_CASTING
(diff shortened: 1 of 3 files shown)


stringdtype/stringdtype/src/casts.c
```diff
@@ -376,44 +372,128 @@ static PyType_Slot s2u_slots[] = {
...
+    ret->name = name;
+    ret->nin = 1;
+    ret->nout = 1;
+    ret->casting = casting;
+    ret->flags = flags;
+    ret->dtypes = dtypes;
+    ret->slots = slots;
+
+    return ret;
+}
```
peytondmurray, line 451: Thanks for this, reduces a lot of duplication.

stringdtype/stringdtype/src/dtype.c
```diff
@@ -393,11 +393,10 @@ init_string_dtype(void)
 
     StringDType.singleton = singleton;
 
-    free(StringDType_DTypeSpec.casts[1]->dtypes);
-    free(StringDType_DTypeSpec.casts[1]);
-    free(StringDType_DTypeSpec.casts[2]->dtypes);
-    free(StringDType_DTypeSpec.casts[2]);
-    free(StringDType_DTypeSpec.casts);
+    for (int i = 0; i < 4; i++) {
```
peytondmurray, line 396: `NULL` is the sentinel here, right? We can just iterate until we hit that, so that we don't have to worry about updating the number of casts here.

peytondmurray (approved):
Looks good, only minor stuff. Thanks!

PR #47, opened 2023-03-08
Add `argmax` `PyArray_ArrFunc`

peytondmurray:
This PR adds support for `argmax`. I've also modified the signature of the `PyArray_CompareFunc` to match the documentation. Additionally, `compare` is now safe against elements with `NULL` data, e.g. `x = np.empty(10); x.sort()` is now safe.

Finally, there was an issue with `isort` in the pre-commit hooks. Autoupdating the hooks resolves this.

### DIFF
--- a/stringdtype/stringdtype/src/dtype.c
+++ b/stringdtype/stringdtype/src/dtype.c
@@ -179,11 +179,28 @@ nonzero(void *data, void *NPY_UNUSED(arr))
 // Implementation of PyArray_CompareFunc.
 // Compares unicode strings by their code points.
 int
-compare_strings(char **a, char **b, PyArrayObject *NPY_UNUSED(arr))
+compare(void *a, void *b, void *NPY_UNUSED(arr))
 {
-    ss *ss_a = (ss *)a;
-    ss *ss_b = (ss *)b;
+    ss *ss_a = NULL;
+    ss *ss_b = NULL;
+    load_string(a, &ss_a);
+    load_string(b, &ss_b);
     return strcmp(ss_a->buf, ss_b->buf);
+}
+
+// PyArray_ArgFunc
+// The max element is the one with the highest unicode code point.
+int
+argmax(void *data, npy_intp n, npy_intp *max_ind, void *arr)
+{
+    ss *dptr = (ss *)data;
+    *max_ind = 0;
+    for (int i = 1; i < n; i++) {
+        if (compare(&dptr[i], &dptr[*max_ind], arr) > 0) {
+            *max_ind = i;
+        }
+    }
+    return 0;
 }
 
 static StringDTypeObject *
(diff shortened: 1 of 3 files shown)


stringdtype/stringdtype/src/dtype.c
```diff
@@ -179,13 +179,32 @@ nonzero(void *data, void *NPY_UNUSED(arr))
...
 int
-compare_strings(char **a, char **b, PyArrayObject *NPY_UNUSED(arr))
+compare(void *a, void *b, void *NPY_UNUSED(arr))
 {
     ss *ss_a = (ss *)a;
     ss *ss_b = (ss *)b;
     return strcmp(ss_a->buf, ss_b->buf);
 }
 
+// PyArray_ArgmaxFunc
```
ngoldbaum, line 189: `PyArray_ArgmaxFunc` isn't a thing, there's a `PyArray_ArgFunc` typedef for both argmin and argmax though.

stringdtype/stringdtype/src/dtype.c
```diff
@@ -179,13 +179,32 @@ nonzero(void *data, void *NPY_UNUSED(arr))
...
+int
+argmax(void *data, npy_intp n, npy_intp *max_ind, void *arr)
+{
+    // Get current index and length to avoid indexing outside of the array
+    PyArrayObject *array = (PyArrayObject *)arr;
+    ss *arr_data = (ss *)PyArray_DATA(array);
+    int start_index = (arr_data - (ss *)data) / PyArray_ITEMSIZE(array);
+
+    *max_ind = start_index;
+    for (int i = start_index; i < n; i++) {
```
ngoldbaum, line 200: The versions of these functions in numpy do:

```
*max_ind = 0
for (int i = 1; i < n; i++)
```

Why is this any different?

ngoldbaum: https://github.com/numpy/numpy/blob/ea0b1708b27e683201285da53c6afda24bc8ba74/numpy/core/src/multiarray/arraytypes.c.src#L3572-L3600

peytondmurray:
I think the tests are failing because the numpy PR on which this PR depends was merged today, and hasn't yet been built in the nightly wheels.

ngoldbaum:
Yeah, it is a little annoying we need to wait on that.

We could bring back https://github.com/numpy/numpy-user-dtypes/pull/4 and build numpy from source. Installing from the wheel was so much faster I dropped it.

peytondmurray:
Ehh, let's just be patient. I hate waiting forever for CI to run, and this only comes up once in a while.

seberg:
Triggered a build.  But maybe we should adjust the numpy `wheels.yml` to just run daily, I feel we have enough users (plus you here) that a daily build isn't a waste of resources anymore.

PR #51, opened 2023-03-29
Add PyArray_FillWithScalarFunc to support arr.fill()

peytondmurray:
This PR adds support for the `arr.fill('blah')` `PyArray_ArrFunc` for the `StringDType`. A simple test was added to check that it works.


peytondmurray:
Yep, looks like this was deprecated - closing.

PR #54, opened 2023-04-04
Add min and max ufuncs

peytondmurray:
This PR adds `min` and `max` ufuncs for the `StringDType`. This is pretty cool because these aren't supported for the fixed length unicode dtypes. This PR adds an `ssset` function, which is similar to `ssdup`, except it doesn't require the output `ss *` to have `len == 0` and `buf == NULL`. Instead, `ssset(in, out)` will copy the input to the output using `realloc` to resize the buffer (or allocating new space if needed). @ngoldbaum I get the sense that this is moving in the right direction but there is enough overlap with `ssdup` that I could also see room for some refactoring here. Let me know what you think.

### DIFF
--- a/stringdtype/stringdtype/src/umath.c
+++ b/stringdtype/stringdtype/src/umath.c
@@ -14,6 +14,101 @@
 #include "umath.h"
 
 static int
+minmax_resolve_descriptors(struct PyArrayMethodObject_tag *NPY_UNUSED(method),
+                           PyArray_DTypeMeta *NPY_UNUSED(dtypes[]),
+                           PyArray_Descr *given_descrs[],
+                           PyArray_Descr *loop_descrs[],
+                           npy_intp *NPY_UNUSED(view_offset))
+{
+    Py_INCREF(given_descrs[0]);
+    loop_descrs[0] = given_descrs[0];
+    Py_INCREF(given_descrs[1]);
+    loop_descrs[1] = given_descrs[1];
+
+    StringDTypeObject *new = new_stringdtype_instance();
+    if (new == NULL) {
+        return -1;
+    }
+    loop_descrs[2] = (PyArray_Descr *)new;
+
+    return NPY_NO_CASTING;
+}
+
+static int
+maximum_strided_loop(PyArrayMethod_Context *NPY_UNUSED(context),
+                     char *const data[], npy_intp const dimensions[],
+                     npy_intp const strides[], NpyAuxData *NPY_UNUSED(auxdata))
+{
+    npy_intp N = dimensions[0];
+    char *in1 = data[0];
+    char *in2 = data[1];
+    char *out = data[2];
+    npy_intp in1_stride = strides[0];
+    npy_intp in2_stride = strides[1];
+    npy_intp out_stride = strides[2];
+
+    while (N--) {
+        if (compare(in1, in2, NULL) > 0) {
+            // Only copy *out* to *in1* if they point to different locations;
+            // for *arr.max()* they point to the same address.
+            if (in1 != out) {
+                ssfree((ss *)out);
+                if (ssdup((ss *)in1, (ss *)out) < 0) {
+                    return -1;
+                }
+            }
+        }
+        else {
+            ssfree((ss *)out);
+            if (ssdup((ss *)in2, (ss *)out) < 0) {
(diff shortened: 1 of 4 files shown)


ngoldbaum:
re the new `ssset` you added, I think I'd like to prefer not using `realloc` and just require calling `ssfree` before `ssdup` if the output buffer might be set already. This is for two reasons, first `realloc` is not guaranteed to return memory to the operating system to re-use unless you pass it a `size` of zero (e.g. equivalent to calling `free`). Second, dealing with realloc failure is tricky, it returns `NULL` to indicate failure, but when it does so the pointer you passed into `realloc` is still valid, so only in the failure case you need to assign the return value of `realloc` to the input pointer.

IMO it's a simpler to just insist that we call `ssfree` first before `ssdup` and `ssnewemptylen` if the output string might be set already.

Will take a closer look at the rest of this PR later.

stringdtype/stringdtype/src/umath.c
```diff
@@ -13,6 +13,95 @@
...
+                           npy_intp *NPY_UNUSED(view_offset))
+{
+    Py_INCREF(given_descrs[0]);
+    loop_descrs[0] = given_descrs[0];
+    Py_INCREF(given_descrs[1]);
+    loop_descrs[1] = given_descrs[1];
+
+    StringDTypeObject *new = new_stringdtype_instance();
+    if (new == NULL) {
+        return (NPY_CASTING)-1;
```
ngoldbaum, line 30: Should return -1 to indicate failure

stringdtype/stringdtype/src/umath.c
```diff
@@ -13,6 +13,95 @@
...
+        if (compare(in1, in2, NULL) > 0) {
+            if (ssset((ss *)in1, (ss *)out) < 0) {
+                return -1;
+            }
+        }
+        else {
+            if (ssset((ss *)in2, (ss *)out) < 0) {
+                return -1;
+            }
+        }
```
ngoldbaum, line 64: I wonder if it would be more performant for reductions to first scan over the input array to find the largest possible string, allocate a temporary string to hold the output, and then instead of doing `malloc/free` over and over for the output just re-use the temporary string buffer. Then when the loop is over, allocate a new `ss` string to hold the result and set it to the output buffer.

I worry that this implementation will lead to memory thrashing in the worst case.

ngoldbaum (commented):
Overall LGTM besides the inline comment and the request to remove `ssset`. Happy to chat about this over a call or whatever later if that would help.

stringdtype/stringdtype/src/umath.c
peytondmurray: Agreed, in fact that's what we do for the `argmin`/`argmax` array funcs. In this case, `maximum_strided_loop` is called for two different purposes which are somewhat in tension here:

1. Inside a reduce function in the case where `arr.max()` is desired
2. When `np.maximum(x, y)` is called, where the return type is an array

In the second case, we want the current behavior. In the first case, we can just loop over the array and find the max without doing `malloc/free` for each element. One thing we could do is detect when we are doing `arr.max` by not doing `malloc/free` in cases where `in1_stride == 0`, since that only arises from the accumulator array in the reduce function.

seberg: I would say the typical approach for this would be to:

1. `get_loop` can attach an `auxdata` with some minimal string size already allocated (can be 0 also).
2. Whenever you need more space, you can have an `unlikely` marked code path to grow that space (`unlikely` may be helpful if it isn't a non-inlined function call anyway).
   * A typical strategy would probably over-allocate a bit since it doesn't hurt (for example, I have something like this in NumPy as `grow_size_and_multiply`, but there are more)
3. auxdata free has to free the space obviously.

This is single pass.  A single length check should be fast (and with overallocation fail rarely).  We just want to avoid that we don't have a lot of unused garbage code in the inner-loop (thus the `unlikely`).

seberg: Wooops, sorry... I suppose for a reduction like this, you want to work "in-place".

So not sure this works super nicely/straight forward.  I suppose this is part of the reason why many string implementations have not just a length but also a "capacity", which allows to overallocate and re-use existing space liberally.

PR #56, opened 2023-04-14
Initial support for missing data

ngoldbaum:
Currently we have two ways of representing an empty string:

```C
ss empty = {0, ""}
ss null = {0, NULL}
```

This PR repurposes the second struct layout to represent missing data, using a new `stringdtype.NA` type. This eliminated the need for the `load_string` helper but necessitated a new `ss_isnull` helper that I've added to the static string header.

Before this PR `np.zeros` depended on the second layout being translated to an empty string by `getitem`, so repurposing that layout to mean missing data means we need additional API surface in the dtype API to specify custom zero-filling logic. See https://github.com/numpy/numpy/pull/23591 for that, this PR depends on that numpy PR being merged first.

To avoid churning through a large number of 1-byte mallocs whenever `np.zeros` is called, I now return a static empty string if a user requests an empty string, and added a check in `ssfree` to avoid freeing that static string.

I've decided it makes sense for `np.empty` to return an array filled with missing data. I could also add additional API in numpy to customize empty-filling if we decide it makes sense to do something else.

I still need to make `NAType` behave more like `NaN` or pandas' `NA` in case someone accesses the NA scalar and uses it in operations with other strings, but I'm punting on that for a future PR.

### DIFF
--- a/stringdtype/stringdtype/src/dtype.c
+++ b/stringdtype/stringdtype/src/dtype.c
@@ -317,17 +317,31 @@ stringdtype_get_clear_loop(void *NPY_UNUSED(traverse_context),
 }
 
 static int
-stringdtype_fill_zero_value(PyArrayObject *arr)
+stringdtype_fill_zero_loop(void *NPY_UNUSED(traverse_context),
+                           PyArray_Descr *NPY_UNUSED(descr), char *data,
+                           npy_intp size, npy_intp stride,
+                           NpyAuxData *NPY_UNUSED(auxdata))
 {
-    char *buf = PyArray_DATA(arr);
-    npy_intp sz = PyArray_SIZE(arr);
-    int elsize = PyArray_DESCR(arr)->elsize;
-
-    for (npy_intp i = 0; i < sz; i++) {
-        if (ssnewlen("", 0, (ss *)(buf + i * elsize)) < 0) {
+    while (size--) {
+        if (ssnewlen("", 0, (ss *)(data)) < 0) {
             return -1;
         }
+        data += stride;
     }
+    return 0;
+}
+
+static int
+stringdtype_get_fill_zero_loop(void *NPY_UNUSED(traverse_context),
+                               PyArray_Descr *NPY_UNUSED(descr),
+                               int NPY_UNUSED(aligned),
+                               npy_intp NPY_UNUSED(fixed_stride),
+                               traverse_loop_function **out_loop,
+                               NpyAuxData **NPY_UNUSED(out_auxdata),
+                               NPY_ARRAYMETHOD_FLAGS *flags)
+{
+    *flags = NPY_METH_NO_FLOATINGPOINT_ERRORS;
+    *out_loop = &stringdtype_fill_zero_loop;
     return 0;
 }
 
@@ -344,7 +358,7 @@ static PyType_Slot StringDType_Slots[] = {
         {NPY_DT_PyArray_ArrFuncs_argmax, &argmax},
         {NPY_DT_PyArray_ArrFuncs_argmin, &argmin},
         {NPY_DT_get_clear_loop, &stringdtype_get_clear_loop},
-        {NPY_DT_fill_zero_value, &stringdtype_fill_zero_value},
+        {NPY_DT_get_fill_zero_loop, &stringdtype_get_fill_zero_loop},
         {0, NULL}};
 
 static PyObject *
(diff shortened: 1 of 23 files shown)


ngoldbaum:
Updated to match updates in the upstream numpy PR. I also bumped the API versions for the other dtypes in anticipation of the upstream numpy getting merged.

The changes to the pre-commit configuration are to resolve issues I was seeing from the existing `build` directory getting generated by an old version of `meson` locally:

```
ERROR: Build directory has been generated with Meson version 0.64.0, which is incompatible with the current version 1.0.1
```

stringdtype/stringdtype/src/dtype.c
```diff
@@ -109,58 +132,74 @@ stringdtype_setitem(StringDTypeObject *NPY_UNUSED(descr), PyObject *obj,
...
-    Py_DECREF(val_obj);
+    if (eq_res < 0) {
+        goto error;
+    }
 
-    if (res == -1) {
-        PyErr_NoMemory();
-        return -1;
+    if (eq_res == 1) {
+        sdata = NULL;
```
peytondmurray, line 155: Can this branch be dropped?

stringdtype/stringdtype/src/dtype.c
```diff
@@ -109,58 +132,74 @@ stringdtype_setitem(StringDTypeObject *NPY_UNUSED(descr), PyObject *obj,
...
-    PyObject *val_obj = PyUnicode_FromStringAndSize(data, len);
-
-    if (val_obj == NULL) {
-        return NULL;
+        char *data = sdata->buf;
+        size_t len = sdata->len;
+        val_obj = PyUnicode_FromStringAndSize(data, len);
+        if (val_obj == NULL) {
+            return NULL;
+        }
```
peytondmurray, line 202: I think this check is unnecessary: `val_obj` gets returned in either case.

stringdtype/stringdtype/src/dtype.c
```diff
@@ -190,10 +229,22 @@ nonzero(void *data, void *NPY_UNUSED(arr))
...
+        // pandas sorts NAs to the end as well
+        // so we follow that behavior here
+        if (!b_is_null) {
+            return 1;
+        }
+        else if (!a_is_null) {
+            return -1;
+        }
+        return 0;
+    }
```
peytondmurray, line 247: I think this can this be simplified, since if either a or b is null we don't need to test the other. 

```c
if (a_is_null) {
    return 1;
}
if (b_is_null) {
    return -1;
}
return strcmp(ss_a->buf, ss_b->buf);
```

stringdtype/stringdtype/src/umath.c
```diff
@@ -125,10 +125,14 @@ string_equal_strided_loop(PyArrayMethod_Context *NPY_UNUSED(context),
     ss *s1 = NULL, *s2 = NULL;
 
     while (N--) {
-        load_string(in1, &s1);
-        load_string(in2, &s2);
-
-        if (s1->len == s2->len && strncmp(s1->buf, s2->buf, s1->len) == 0) {
+        s1 = (ss *)in1;
+        s2 = (ss *)in2;
+        if (ss_isnull(s1) || ss_isnull(s2)) {
```
peytondmurray, line 130: I was thinking that it might be good if `NA` values were considered equal, but it looks like numpy doesn't consider `nan` values equal either, so this seems consistent with that. :+1:

peytondmurray (approved):
Had a few minor suggestions - thanks for adding this!

The only thing that might be good to add is a test to make sure that NA-values get sorted to the end. It would also be fine to add this later too.

stringdtype/stringdtype/src/dtype.c
ngoldbaum: Nope, I'll add a comment explaining what it does though since it's clearly not obvious :)

stringdtype/stringdtype/src/dtype.c
ngoldbaum: I'd prefer to keep this just in case someone adds any further logic after this that might use `val_obj`.

ngoldbaum:
Darn looks like the new wheel hasn't gotten uploaded yet. @seberg any chance you can kick the numpy wheel builder?

ngoldbaum:
Merging to unbreak the tests. If anyone has additional comments I'm happy to address in a followup.

stringdtype/stringdtype/src/umath.c
seberg: Pandas would consider `NA == NA -> NA` but that requires an NA boolean.  I suspect that would be the right thing in principle, though.

PR #57, opened 2023-04-14
Add an add ufunc

peytondmurray:
This PR adds an `add` ufunc which concatenates strings element-wise, and a test to make sure it works the same as python string concatenation works.

### DIFF
--- a/stringdtype/stringdtype/src/umath.c
+++ b/stringdtype/stringdtype/src/umath.c
@@ -14,24 +14,41 @@
 #include "umath.h"
 
 static int
-minmax_resolve_descriptors(struct PyArrayMethodObject_tag *NPY_UNUSED(method),
-                           PyArray_DTypeMeta *NPY_UNUSED(dtypes[]),
-                           PyArray_Descr *given_descrs[],
-                           PyArray_Descr *loop_descrs[],
-                           npy_intp *NPY_UNUSED(view_offset))
+add_strided_loop(PyArrayMethod_Context *NPY_UNUSED(context),
+                 char *const data[], npy_intp const dimensions[],
+                 npy_intp const strides[], NpyAuxData *NPY_UNUSED(auxdata))
 {
-    Py_INCREF(given_descrs[0]);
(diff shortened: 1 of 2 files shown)


stringdtype/stringdtype/src/umath.c
```diff
@@ -13,6 +13,65 @@
...
+                        npy_intp *NPY_UNUSED(view_offset))
+{
+    Py_INCREF(given_descrs[0]);
+    loop_descrs[0] = given_descrs[0];
+    Py_INCREF(given_descrs[1]);
+    loop_descrs[1] = given_descrs[1];
+
+    StringDTypeObject *new = new_stringdtype_instance();
+    if (new == NULL) {
+        return -1;
```
seberg, line 30: The design is currently to ask you to clean up here.  So its to put it this first (IMO).

But you should not need a resolve descriptors at all, since the dtype isn't parametric?

peytondmurray: You're totally right, I moved this to the top.

After reading more, I modified `init_ufunc` to accept `NULL` as a way of indicating that the ufunc uses `default_resolve_descriptors`. I was able to eliminate another place where this was happening as well for the `min`/`max` ufuncs.

stringdtype/stringdtype/src/umath.c
```diff
@@ -367,14 +393,21 @@ init_ufuncs(void)
 
     PyArray_DTypeMeta *minmax_dtypes[] = {&StringDType, &StringDType,
                                           &StringDType};
-    if (init_ufunc(numpy, "maximum", minmax_dtypes,
-                   &minmax_resolve_descriptors, &maximum_strided_loop,
```
ngoldbaum, line 371: Can't you delete `minmax_resolve_descriptors` now too?

peytondmurray: Wait, didn't I? I think it's been taken out.

ngoldbaum: Sorry I have code blindness sometimes 😭

PR #58, opened 2023-04-27
Make it possible for users to use a custom NA value

ngoldbaum:
This makes it possible to create a `StringDType` instance with a missing data sentinel supplied by the user. I'm going to use this in my pandas fork so that the dtype handles inserting `pd.NA` for the pandas extension dtype I'm writing.

I'm putting this in now because it works, however there's a design issue in that `discover_descriptor_from_pyobject` can't possibly work correctly. It *would* work if I could specialize at the level of the dtype class, since that function is supplied a reference to the  the dtype class. That approach would also let me avoid making the dtype parametric, which would simplify things a bit.

I'm going to try implementing this again, storing the na object in the DType class's `tp_dict`. If that turns out to have worse performance we can move forward with this approach.

### DIFF
--- a/stringdtype/stringdtype/src/casts.c
+++ b/stringdtype/stringdtype/src/casts.c
@@ -22,11 +22,8 @@ string_to_string_resolve_descriptors(PyObject *NPY_UNUSED(self),
                                      npy_intp *view_offset)
 {
     if (given_descrs[1] == NULL) {
-        StringDTypeObject *new = new_stringdtype_instance();
-        if (new == NULL) {
-            return (NPY_CASTING)-1;
-        }
-        loop_descrs[1] = (PyArray_Descr *)new;
+        Py_INCREF(given_descrs[0]);
+        loop_descrs[1] = given_descrs[0];
     }
     else {
         Py_INCREF(given_descrs[1]);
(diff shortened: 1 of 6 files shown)


seberg:
NA is tricky with dtype discovery, since in practice a generic NA has no dtype.  I think the only way to solve this, maybe be to add support for this in some form.  That could be in form of an abstract DType.  But NumPy currently (for simplicity reasons) wants to work with `dtype` instances for discovery in `np.array()`.

We could special case a `NoDType/NADType`, which actually has no descriptor instance at all (just remains NULL and we special case that).
The maybe better path is, to finally accept that we will need abstract instances anyway, since all those really need is a simple scheme like `itemsize == -1` and array creation rejects it already (plus ability to convert it to a concrete version).

(the problem is that `np.array([NA, NA])` should or should not work by defaulting to float64, but `np.array([NA, string])` should really manage to figure out the string dtype.

peytondmurray (approved):
Thanks for this! :tada:

peytondmurray:
After talking with @ngoldbaum about this we will come back to the issue of making dtypes heap-allocated later on, and for now merge this as is.

PR #60, opened 2023-04-29
[WIP] Add a multiply ufunc for the StringDType

peytondmurray:
This PR adds a multiply ufunc for the `StringDType`.

This is a WIP, I need to make it work with the most recent commit that adds user-defined NA values.

### DIFF
--- a/stringdtype/stringdtype/src/umath.c
+++ b/stringdtype/stringdtype/src/umath.c
@@ -1,5 +1,7 @@
 #include <Python.h>
 
+#include "object.h"
+
 #define PY_ARRAY_UNIQUE_SYMBOL stringdtype_ARRAY_API
 #define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
 #define NO_IMPORT_ARRAY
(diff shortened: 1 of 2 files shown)


ngoldbaum:
The seg fault is happening because `binary_resolve_descriptors` expects both of the `given_descrs` to be `StringDType` instances, but that's not true for the multiply test. You probably need to write a different `resolve_descriptors` for `multiply` specifically that doesn't make that assumption. In fact, I think it should probably raise an error if the non-string descriptor doesn't represent an integer.

ngoldbaum:
still WIP?

peytondmurray:
Yep, for a bit longer - still debugging a segfault:

```python
import numpy as np

x = np.array(["abc", "def", "ghi", "A¢☃€ 😊", "Abc", "DEF"], dtype=StringDType())

print(np.multiply(3, x)) #  <- this one segfaults, still working on it
print(np.multiply(x, 3))
```

stringdtype/stringdtype/src/umath.c
```diff
@@ -46,6 +48,73 @@ binary_resolve_descriptors(struct PyArrayMethodObject_tag *NPY_UNUSED(method),
...
+        struct PyArrayMethodObject_tag *NPY_UNUSED(method),
+        PyArray_DTypeMeta *NPY_UNUSED(dtypes[]), PyArray_Descr *given_descrs[],
+        PyArray_Descr *loop_descrs[], npy_intp *NPY_UNUSED(view_offset))
+{
+    Py_INCREF(given_descrs[0]);
+    loop_descrs[0] = given_descrs[0];
+    Py_INCREF(given_descrs[1]);
+    loop_descrs[1] = given_descrs[1];
+    Py_INCREF(given_descrs[0]);
+    loop_descrs[2] = given_descrs[0];
```
seberg, line 62: This isn't right.

ngoldbaum:
Reissuing from my fork with some improvements

PR #69, opened 2023-06-01
Fix equality comparisons with object arrays

ngoldbaum:
This adds a cast from object to string, mostly cribbed off of the generic object to any cast in numpy. I can't use that cast directly for this purpose because ufuncs default to `same_kind` casting, but the generic object to any cast is defined as an unsafe cast. For strings specifically, it's a safe cast because string arrays can only be constructed from scalars that are already python strings and will raise a `TypeError` otherwise. The version of the cast added in this PR is defined as a safe cast, it also uses `stringdtype_setitem` directly instead of going through PyArray_Pack, which  isn't public in the numpy C API.

In addition to the cast, this adds promoters for the `equal` and `not_equal` ufuncs, so object string arrays get promoted to stringdtype and then we can use the string equals ufunc implementation.

If there's a way to do this without copy/pasting the object to any ufunc from numpy's internals I'd love to hear it.

### DIFF
--- a/stringdtype/stringdtype/src/casts.c
+++ b/stringdtype/stringdtype/src/casts.c
@@ -449,6 +449,124 @@ static PyType_Slot s2b_slots[] = {
 
 static char *s2b_name = "cast_StringDType_to_Bool";
 
+// object to string
+
+typedef struct {
+    NpyAuxData base;
+    PyArray_Descr *descr;
+    int move_references;
+} _object_to_string_auxdata;
+
+static void
+_object_to_string_auxdata_free(NpyAuxData *auxdata)
+{
+    _object_to_string_auxdata *data = (_object_to_string_auxdata *)auxdata;
+    Py_DECREF(data->descr);
+    PyMem_Free(data);
+}
+
+static NpyAuxData *
+_object_to_string_auxdata_clone(NpyAuxData *data)
+{
+    _object_to_string_auxdata *res = PyMem_Malloc(sizeof(*res));
+    if (res == NULL) {
+        return NULL;
+    }
+    memcpy(res, data, sizeof(*res));
+    Py_INCREF(res->descr);
+    return (NpyAuxData *)res;
+}
+
+static int
+object_to_string_strided_loop(PyArrayMethod_Context *NPY_UNUSED(context),
+                              char *const *args, const npy_intp *dimensions,
+                              const npy_intp *strides, NpyAuxData *auxdata)
+{
+    npy_intp N = dimensions[0];
+    char *src = args[0], *dst = args[1];
+    npy_intp src_stride = strides[0], dst_stride = strides[1];
+    _object_to_string_auxdata *data = (_object_to_string_auxdata *)auxdata;
+
+    PyObject *src_ref;
+
+    while (N > 0) {
+        memcpy(&src_ref, src, sizeof(src_ref));
+        if (stringdtype_setitem((StringDTypeObject *)(data->descr),
+                                src_ref ? src_ref : Py_None,
+                                (void *)dst) < 0) {
+            return -1;
+        }
+
+        if (data->move_references && src_ref != NULL) {
+            Py_DECREF(src_ref);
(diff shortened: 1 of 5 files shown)


ngoldbaum:
The most recent push enables the rest of the comparison operators and adds promoters for those operators so comparison with unicode and object string arrays will work as expected.

The strided loops for all the comparison operators are almost identical. I don't think it's possible to use a macro to avoid this. It is possible to use C++ for this file to avoid the copy/paste on all the strided loops, but I don't think the cost of the repeated code is higher than the cost of needing to worry about C++ build issues. We can come back to that choice if I find myself writing a lot more code that would be aided by templates.

seberg:
Yeah, we should make the `PyArray_Pack` public, maybe with a fallback to setitem on older versions (for 2.0 if we have a compat layer)...

I am not sure sure I follow the approach here, why would you add the `string -> object` cast?  Isn't an unsafe cast correct there (you can't cast quite everything to a string).  Also, isn't casting incorrect in the sense of comparison since `None != "None"`?

I.e. I am a bit confused why the comparison shouldn't go through `object` dtype and already work?  Is that about speed, or some other difficulty in pandas?

ngoldbaum:
> why would you add the string -> object cast?

So the basic issue is that without this PR we have this situation, which is really bad because we're getting incorrect answers:

```
In [1]: sarr
Out[1]: array(['a', 'b', 'c'], dtype=StringDType())

In [2]: oarr
Out[2]: array(['a', 'b', 'c'], dtype=object)

In [3]: sarr == oarr
Out[3]: array([False, False, False])
```

I'm not totally sure why we're getting incorrect answers here and not an error, because:

```
In [6]: np.equal(sarr, oarr)
---------------------------------------------------------------------------
UFuncTypeError                            Traceback (most recent call last)
Cell In[6], line 1
----> 1 np.equal(sarr, oarr)

UFuncTypeError: ufunc 'equal' did not contain a loop with signature matching types (<class 'stringdtype.StringDType'>, <class 'numpy.dtypes.ObjectDType'>) -> None
```

Maybe there's a fallback path somewhere that returns false if a ufunc doesn't have a matching loop?

There already is a safe cast from string to object and it didn't occur to me while I was working on this PR that I can add a promoter that converts the inputs to object dtype. Right now there's only one promoter implementation in stringdtype and it just converts all the arguments to stringdtype, which made me think I *needed* to cast to stringdtype. I think I'm going to try following that approach so I don't need to go so hard against the grain here and also probably user expectations.

To justify the approach in this PR a little bit more, right now the cast from object to `stringdtype` specifically is safe because stringdtype rejects anything that's not already a string and doesn't try to coerce a generic python object via `__str__`, e.g.:

```
In [7]: sarr[1] = None
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[7], line 1
----> 1 sarr[1] = None

In [8]: np.array([1, 2, None], dtype=StringDType())
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[8], line 1
----> 1 np.array([1, 2, None], dtype=StringDType())

TypeError: Can only store String text in a StringDType array.
```

Whether or not that's a good design or aligns with user expectations is another story, but if we can avoid forcing that choice on ourselves let's avoid it.

seberg:
> I'm not totally sure why we're getting incorrect answers here and not an error, because:

Yes, `==` and `!=` convert the error.  That was the solution to that infinite FutureWarning.

> UFuncTypeError: ufunc 'equal' did not contain a loop with signature matching types (<class 'stringdtype.StringDType'>, <class 'numpy.dtypes.ObjectDType'>) -> None

Well, that is the issue then... The promotion should work, you can promote string to object after all.  Probably, that just shouldn't surprise me, unless you don't define the string -> object cast as always safe?

seberg:
Yeah, maybe we could use "safe" also in cases where it's safe if it works but can raise.

ngoldbaum:
Closing this and opening a new PR since the original description no longer applies.

PR #71, opened 2023-06-15
simplify stringdtype get_value implementation

ngoldbaum:
Up until now I've been coercing `None`, `np.nan`, and float('nan') to the dtype's na value in the `get_value` function. This means that `setitem` and array creation will silently coerce these values to the na value. This made things a little easier for me for pandas support, but today I've figured out how to refactor my pandas changes to avoid this in stringdtype.

This PR makes it so only the na value for that dtype gets special treatment, all other values are either already strings or get coerced to string.

### DIFF
--- a/stringdtype/stringdtype/src/dtype.c
+++ b/stringdtype/stringdtype/src/dtype.c
@@ -81,7 +81,7 @@ get_value(PyObject *scalar, StringDType_type *cls)
     else if (!((scalar_type == &PyUnicode_Type) ||
                (scalar_type == expected_scalar_type))) {
         // attempt to coerce to str
-        scalar = PyObject_CallOneArg((PyObject *)&PyUnicode_Type, scalar);
+        scalar = PyObject_Str(scalar);
         if (scalar == NULL) {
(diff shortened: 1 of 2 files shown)


stringdtype/stringdtype/src/dtype.c
```diff
@@ -63,52 +63,32 @@ common_dtype(PyArray_DTypeMeta *cls, PyArray_DTypeMeta *other)
...
-                    PyExc_TypeError,
-                    "Can only store UTF8 text in a StringDType array.");
+    if (scalar == cls->na_object) {
+        Py_INCREF(scalar);
+        return scalar;
+    }
+    else if (!((scalar_type == &PyUnicode_Type) ||
+               (scalar_type == expected_scalar_type))) {
+        // attempt to coerce to str
+        scalar = PyObject_CallOneArg((PyObject *)&PyUnicode_Type, scalar);
```
seberg, line 84: Nice simplification!  Just thought I would comment that `PyObject_Str` might be easier to read and is presumably equivalent.  But look good either way.

PR #72, opened 2023-06-20
Add round-trip casts between string and (u)int dtypes

ngoldbaum:
Making a PR to solicit feedback before I do this again for the float dtypes.

I could have done this without using the python C API but this approach seemed less finicky than dealing with serializing ints as strings in a portable way in C. FWIW, numpy currently does the equivalent cast for the unicode dtype by going through numpy scalars. In principle in the future we could drop the GIL requirement and do this in a more low-level way.

There's a lot of copy/paste here. It's very possible we've reached or are long past the point where writing this in C++ would be a good idea. I guess if this code is going to end up living in numpy eventually it's not problematic to add a build dependency on a C++ toolchain.

### DIFF
--- a/asciidtype/asciidtype/src/asciidtype_main.c
+++ b/asciidtype/asciidtype/src/asciidtype_main.c
@@ -22,7 +22,7 @@ PyInit__asciidtype_main(void)
         return NULL;
     }
 
-    if (import_experimental_dtype_api(11) < 0) {
+    if (import_experimental_dtype_api(12) < 0) {
         return NULL;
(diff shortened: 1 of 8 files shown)


ngoldbaum:
> There's a lot of copy/paste here.

It turns out I can avoid almost all of the copy/paste using preprocessor macros, see latest commit.

ngoldbaum:
I'm happy with the state of this now, if anyone has any comments or suggestions I'll take care of it in a followup.

PR #73, opened 2023-06-22
Add casts for integer dtypes that don't have a sized alias

ngoldbaum:
I've also used macros in a couple of places to reduce the amount of copy/paste in this file.


ngoldbaum:
I think I'm pretty happy with how the integer casts turned out using macros. I'm going to adopt that approach in a followup with the float dtypes.

PR #92, opened 2023-10-23
Arena allocator refactor

ngoldbaum:
This refactors the implementation to use an arena allocator.

The `StringDTypeObject` struct gains two new fields: `int array_owned` and `npy_string_allocator *allocator`. The `array_owned` flag is used to indicate if an array owns the descriptor or not and the `allocator` is used to mediate heap memory access and allocation.

The main new feature here is that newly created arrays store data in a contiguous buffer that lives inside an `npy_string_arena` struct (that itself is an attribute of the `allocator` struct). There's only one arena per dtype instance - as declared with the `NPY_DT_UNIQUE_ARRAY_DESCRIPTOR` flag on the DType with bookeeping managed via the `array_owned` flag.

The allocator and arena structs look like this:

```C

typedef struct npy_string_arena {
    size_t cursor;
    size_t size;
    char *buffer;
} npy_string_arena;

typedef void *(*npy_string_malloc_func)(size_t size);
typedef void (*npy_string_free_func)(void *ptr);
typedef void *(*npy_string_realloc_func)(void *ptr, size_t size);

struct npy_string_allocator {
    npy_string_malloc_func malloc;
    npy_string_free_func free;
    npy_string_realloc_func realloc;
    npy_string_arena arena;
};
```

Heap-allocated string data lives in `allocator->arena->buffer`. Internally the buffer works by prepending each string allocation with a field indicating the size of the string. Right now it's always a `size_t`, but that adds a lot of overhead for small strings (alignment makes it worse) so I'd like to add a mode where small heap strings have lengths stored in a `char` or `unsigned char`.

Another nice aspect of this is now packed heap strings that are in the arena don't store a memory address directly, instead they now store an offset into the arena.

I still have it set up so that if you modify an existing string and the modification would exceed the available space to store that string, the modified string is allocated on the heap using malloc/free directly.

The bookeeping for this is mediated via the new `NPY_STRING_ARENA_FREED` and `NPY_STRING_ON_HEAP` flags. The former indicates a packed string that is freed but not yet overwritten. The latter indicates a packed string that corresponds to a string allocated directly on the heap, outside the arena buffer. Unfortunately this uses up the two unused flag bits we had available, although we still have four flags available for non-short strings.

### DIFF
--- a/stringdtype/stringdtype/src/casts.c
+++ b/stringdtype/stringdtype/src/casts.c
@@ -72,13 +72,12 @@ string_to_string(PyArrayMethod_Context *context, char *const data[],
                  npy_intp const dimensions[], npy_intp const strides[],
                  NpyAuxData *NPY_UNUSED(auxdata))
 {
-    StringDTypeObject *in_descr =
-            ((StringDTypeObject *)context->descriptors[0]);
-    StringDTypeObject *out_descr =
-            ((StringDTypeObject *)context->descriptors[1]);
-    int in_hasnull = in_descr->na_object != NULL;
-    int out_hasnull = out_descr->na_object != NULL;
-    const npy_static_string *in_na_name = &in_descr->na_name;
+    StringDTypeObject *idescr = (StringDTypeObject *)context->descriptors[0];
+    StringDTypeObject *odescr = (StringDTypeObject *)context->descriptors[1];
+    npy_string_allocator *allocator = odescr->allocator;
+    int in_hasnull = idescr->na_object != NULL;
+    int out_hasnull = odescr->na_object != NULL;
+    const npy_static_string *in_na_name = &idescr->na_name;
     npy_intp N = dimensions[0];
     char *in = data[0];
     char *out = data[1];
@@ -89,17 +88,17 @@ string_to_string(PyArrayMethod_Context *context, char *const data[],
         const npy_packed_static_string *s = (npy_packed_static_string *)in;
         npy_packed_static_string *os = (npy_packed_static_string *)out;
         if (in != out) {
-            npy_string_free(os);
+            npy_string_free(os, allocator);
             if (in_hasnull && !out_hasnull && npy_string_isnull(s)) {
                 // lossy but this is an unsafe cast so this is OK
-                if (npy_string_newsize(in_na_name->buf, in_na_name->size, os) <
-                    0) {
+                if (npy_string_newsize(
+                            in_na_name->buf, in_na_name->size, os, allocator) < 0) {
                     gil_error(PyExc_MemoryError,
                               "Failed to allocate string in string to string "
                               "cast.");
                 }
             }
-            else if (npy_string_dup(s, os) < 0) {
+            else if (npy_string_dup(s, os, allocator) < 0) {
                 gil_error(PyExc_MemoryError, "npy_string_dup failed");
                 return -1;
             }
@@ -211,6 +210,8 @@ unicode_to_string(PyArrayMethod_Context *context, char *const data[],
                   NpyAuxData *NPY_UNUSED(auxdata))
 {
     PyArray_Descr **descrs = context->descriptors;
+    npy_string_allocator *allocator =
+            ((StringDTypeObject *)descrs[1])->allocator;
     long max_in_size = (descrs[0]->elsize) / 4;
 
     npy_intp N = dimensions[0];
@@ -230,8 +231,8 @@ unicode_to_string(PyArrayMethod_Context *context, char *const data[],
             return -1;
         }
         npy_packed_static_string *out_pss = (npy_packed_static_string *)out;
-        npy_string_free(out_pss);
-        if (npy_string_newemptysize(out_num_bytes, out_pss) < 0) {
+        npy_string_free(out_pss, allocator);
+        if (npy_string_newemptysize(out_num_bytes, out_pss, allocator) < 0) {
             gil_error(PyExc_MemoryError,
                       "Failed to allocate string in unicode to string cast");
             return -1;
@@ -493,7 +494,7 @@ static char *s2b_name = "cast_StringDType_to_Bool";
 // bool to string
 
 static int
-bool_to_string(PyArrayMethod_Context *NPY_UNUSED(context), char *const data[],
+bool_to_string(PyArrayMethod_Context *context, char *const data[],
                npy_intp const dimensions[], npy_intp const strides[],
                NpyAuxData *NPY_UNUSED(auxdata))
 {
@@ -504,9 +505,12 @@ bool_to_string(PyArrayMethod_Context *NPY_UNUSED(context), char *const data[],
     npy_intp in_stride = strides[0];
     npy_intp out_stride = strides[1];
 
+    StringDTypeObject *descr = (StringDTypeObject *)context->descriptors[1];
+    npy_string_allocator *allocator = descr->allocator;
+
     while (N--) {
         npy_packed_static_string *out_pss = (npy_packed_static_string *)out;
-        npy_string_free(out_pss);
+        npy_string_free(out_pss, allocator);
         char *ret_val = NULL;
         size_t size = 0;
         if ((npy_bool)(*in) == 1) {
@@ -522,7 +526,7 @@ bool_to_string(PyArrayMethod_Context *NPY_UNUSED(context), char *const data[],
                       "invalid value encountered in bool to string cast");
             return -1;
         }
-        if (npy_string_newsize(ret_val, size, out_pss) < 0) {
+        if (npy_string_newsize(ret_val, size, out_pss, allocator) < 0) {
             // execution should never get here because this will be a small
             // string on all platforms
             gil_error(PyExc_MemoryError,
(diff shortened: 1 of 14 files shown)


stringdtype/stringdtype/src/casts.c
```diff
@@ -42,8 +42,7 @@ string_to_string_resolve_descriptors(PyObject *NPY_UNUSED(self),
                                      npy_intp *view_offset)
 {
     if (given_descrs[1] == NULL) {
-        Py_INCREF(given_descrs[0]);
-        loop_descrs[1] = given_descrs[0];
+        loop_descrs[1] = stringdtype_get_unique_array_descr(given_descrs[0]);
```
seberg, line 45: Mainly thinking.  This situation is interesting because we have 2/3 different situations:
* `arr[...] = arr2` assignments, where the result dtype is "fixed"
* `arr.astype(arr2.dtype)` where the result dtype is requested, but we need a new one.
* `arr.astype(StringDType)` where the result dtype is `NULL`.  I think in this case we can assume a new array is being created.

In the end, we just don't know if we are dealing with an existing allocation or a new one!  One additional caveat is that saying that `view_offset=0` seems only reasonable if we are actually dealing with an existing allocation.

Thoughts (maybe random):
* We could make sure `.astype(dtype)` calls the `get_unique_descr` function or similar?  But:
  * `np.empty()`, etc. it would be nice if the array creation could take care of setting "owned" (and creating a new instance).
  * If `astype()` creates the unique dtype (when passed in), it could indicate to the array creation to use the descriptor verbatim in principle, or would it be necessary to flag that the descriptor is "unowned" and the array creation can just set it to "owned"?
* I was thinking about just using "not owned" but keeping the allocator might work (and then swapping it out).  But that breaks `view_offset=0` semantics, I think.  (i.e. figuring this out in hind-sight).

Have to think about it more, sorry its fuzzy.  Maybe we can also discuss async/sync, but maybe you have some ideas based on this.

ngoldbaum: Hmm, I was under the (mistaken?) impression that we only entered the `given_descrs[1] == NULL` branch if someone does something like `astype(StringDType)`.

I'm having some trouble coming up with examples where this happens. Both scripts in the details blocks below go into the other branch in gdb.

> arr[...] = arr2 assignments, where the result dtype is "fixed"

<details>

```python
import numpy as np

arr = np.array(["hello" * 25, "world" * 25] * 25, dtype=StringDType())

dt = arr.dtype

arr2 = np.array(["foo" * 25, "bar" * 25] * 25, dtype=StringDType())
arr[...] = arr2

print(arr)
assert dt is arr.dtype
```

</details>

> arr.astype(arr2.dtype) where the result dtype is requested, but we need a new one.

<details>

```python

import numpy as np
from numpy._core._multiarray_umath import _get_sfloat_dtype

arr = np.array(["hello" * 25, "world" * 25] * 25, dtype=StringDType(na_object=np.nan))
arr2 = np.array(["hello"*10, "world"*10], dtype=StringDType)
dt = arr2.dtype
arr2 = arr.astype(arr2.dtype)

print(arr2)
assert dt is not arr2.dtype

```

</details>

seberg: Right, the `NULL` case is probably not very interesting.  And yes both of those cases look identical to resolve dtypes.  Part of the problem may be that we need to distinguish the two cases.
But, I think I thought in the wrong direction.  This may be a problem, but `astype()` currently works like this:
```
# cast user provided dtype, which uses `resolve_descriptors`
new_dtype = cast_dtype(arr.dtype, dtype)
result = np.empty_like(arr, dtype=new_dtype)
result[...] = arr
```
So we we (may) call `resolve_descriptors` twice, but it gets the right descriptor.  And this is not a problem (resolve dtypes always gets the final dtype if it gets one).
But, it may still be important to keep in mind, that in general we have to ensure this.

So back to the multi-step cast:
* We are casting `orig_string -> unicode` and then insert `orig_string -> make_unique(orig_string)` before that.
  * This is because we are creating a new buffer (for silly reasons, but OK).
  * I find it a bit awkward that the second cast works with a different instance then the first.  This works out because `make_unique()` shouldn't do anything weird in this case (even if you broaden its meaning, `orig_string` was already attached to an array after all).  So this is a requirement here that `make_unique` must return something that is OK to read from with the original dtype.  That dtype must have had `make_unique()` called on it, so this is a requirement that `make_unique(make_unique(dtype))` and `make_unique(dtype)` are ABI compatible for *reading*.
    * While we might know we will need a buffer, unfortunately we don't know whether we need a multi-step cast anyway, so I think we cannot move the `make_unique()` to before setting up the first cast :(.
* I am worried a bit because having to insert it here for a buffer creation means that we probably have to do the same thing in other places also.  There shouldn't be a huge number of such buffers being set up, so it isn't a catastrophy.
  * With adapting `asdtype()` it might be that any `resolve_descriptors()` call should never have the identical input/output descriptor, might be a way to add an assert at least?
  * Unfortunately, I am not sure we could use this to decide to return a new one in `resolve_descriptors()` for convenience (to some paths at leasts).  Right now, it enforces identity for same DType casts for the sake of it only.  So we could allow this (and then use the returned one for the buffer), however it would mean that we cannot allow within DType casts to have NumPy set up a multi-step cast behavior for them in the future.  (Which could be neat for complicated unit casts.)

OK, too long... in the end, I am suspecting we really just need it and the `make_unique` constraint is fine in practice.  We *could* yet replace the dtype for the `get_loop()` call, which moves the constraint that `resolve_descriptors()` cannot assume that `make_unique()` won't mutate descriptors before the `get_loop()` call, which is maybe clearer.
The problem with buffer set up might just be minor, but we will have to take a lot more care about cases where we assume dtypes will always be the same.  Two such examples:
* The (fast)take implementation assumes the dtype is identical.  If we assume that make-unique doesn't bother *reading*, all that is needed is inserting the call (or fetching the actual descriptor from the array, but there is the subtlty that our `S0` is barfed as it gets replaced with `S1` by default, so that might need fixing at the same time.)
* The sorting buffer setup needs the call.

seberg: Hmmm, the `unicode -> string` cast may insert a step also afterwards.  That seems wrong in the other PR?  Because I think you made it `string -> make_unique(string)`, but we must not mutate the result dtype there since it may already be attached to a NumPy array?

Maybe it is correct to move the make unique ahead of time after all?
* If we know we *must* use a buffer, just make the dtype unique for that step.  But for deciding whether to *chain* we must keep using the original dtypes of course!
  * If we must chain anyway due to casting, it shouldn't matter whether the input was "made unique", we replace it again with something that is unique anyway.
  * If the cast can work with the dtype directly, we will still notice the necessary copy step, since we killed the dtype identity ahead of time.

In the end, every time we know we are creating a new buffer, it seems like we want to create a new unique dtype.

I am also wondering whether we should actually indicate if a dtype is made unique for buffering purposes or at array creation.

ngoldbaum: Good call on `sort` and `fasttake` - the existing tests I had for those didn't create any non-short strings so I didn't catch those cases.

ngoldbaum:
Tests are passing again. 

I'm very aware this is an unreviewably large PR. @seberg do you have any desire to look over these changes at a high level?

I realized that I *do* actually need to store the allocation size, so I need to come back to the over-allocation issues related to that. I'd also like to figure out if it's possible to do a final realloc after the array buffer is fully initialized to resize the sidecar buffer to only what's needed. I'd like to do that in a followup, though.

For now I've xfailed the test that's failing due to a ufunc error because upstream changes in numpy to handle python integers in ufuncs. I'll also come back to this later or fix it upstream in numpy.

seberg:
Happy if you just merge.  I can try to have a look over soonish, but overall, it's probably just as well to continue.  Do you have anything specific you would like me to look at closer?

ngoldbaum:
OK, let's pull this in then. I think we've hashed out the bits I wanted you to look carefully at already and I think the high-level design is fine and more or less matches what's in the NEP already.

seberg:
About a potential array-is-filled hook.  I don't mind the idea in general, since it could be useful.  But do wonder if we need it here.  Because we may mutate the array anyway or use it for buffers.  OTOH, making an immutable string array seems like it could be worthwhile and that may just be the right approach.

I also slightly wonder if traversal during the GC might actually be better point where to do memory management book-keeping (if necessary).

seberg:
We should have a traversal function for GC traversal.  However, I will note that NumPy simply doesn't do GC traversal currently.  (which I hate, but no review meant fixing it never went anywhere, although now with the new traversals adding it might also be a bit saner.)

ngoldbaum:
I guess you're referring to https://github.com/numpy/numpy/pull/15065? Reviving that might be more than I can take on right now but let's chat a bit about this at today's triage meeting if you're going to be there. Something like this is needed if we want to move to heap types, I think.

seberg:
It's actually not hard to do now, TBH.  So we either one or both of:
* Jus talways set the flag and don't try to be smart
* Ignore any reference cycles that may occur due to the dtype itself.

That PR wasn't even particlarly tricky, just ugly and with that subtlety above that I don't have a clear answer to.  Plus a large chunk of "is it slow?", which I honestly don't want to even start answering, because it also is a clear bug.

PR #93, opened 2023-11-01
Add a medium length string mode

ngoldbaum:
This adds a "medium string" mode where the sizes of arena allocations are stored using a `char` instead of a `size_t`, saving 8 bytes per string for strings with lengths between 16 and 255 bytes. I modified the tests so they use longer strings more often which exercises all the different memory allocation modes we're now supporting.

I also dropped the attempt to store the `size_t` sizes in aligned locations because it adds a bunch of bookkeeping headaches. If it turns out the `memcpy` to unaligned addresses is slow we can come back to this.


ngoldbaum:
I'm glad I have ASAN running in the tests here, it noticed a serious flaw in the first version of this.

PR #95, opened 2023-11-16
Updates to match the NEP

ngoldbaum:
This is mostly cosmetic changes to match some of the naming I arrived at in the NEP. I also fixed a reference counting issue in `stringdtype_setitem` that caused a memory leak I hadn't noticed until now.

I also updated the README. @rgommers is there any chance you can try this out in a fresh python environment and let me know if you run into issues? I'd like to include a call for testing the dtype in my message to the numpy mailing list about the NEP updates.

### DIFF
--- a/stringdtype/README.md
+++ b/stringdtype/README.md
@@ -1,16 +1,17 @@
 # A dtype that stores pointers to strings
 
-This is a simple proof-of-concept dtype using the (as of early 2023) experimental
-[new dtype
-implementation](https://numpy.org/neps/nep-0041-improved-dtype-support.html) in
-NumPy.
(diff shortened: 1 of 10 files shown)


ngoldbaum:
@seberg in the last few commits I refactored it so `npy_packed_static_string` is an opaque struct. I also added `NpyString_pack` to mirror `NpyString_load` and used it in a number of places where it made sense to do so.

The only slightly tricky thing is the dtype implementation needs to know the struct size and alignment to set the elsize, so I defined a macro that handles that. The macro and the layout just need to be kept in sync if they're ever changed.

ngoldbaum:
OK, this now fully matches what I have in the latest version of the NEP. Pulling this in.

PR #96, opened 2023-11-21
add a promoter for multiplying with a python int

ngoldbaum:
This fixes multiplication with a python int by adding a promoter for that case. @seberg do you foresee any issues with promoting all pyints to int64?

There's also some refactoring of the existing ufunc and promoter setup to delete unnecessary code or code that's repeated unnecessarily, let me know if you need any clarifications.


seberg:
> This fixes multiplication with a python int by adding a promoter for that case. @seberg do you foresee any issues with promoting all pyints to int64?

No, that is totally safe unless you wanted to support gigantic ones (which seems silly).

PR #97, opened 2024-07-16
Update for numpy 2.0

ngoldbaum:
With this all the dtypes build and the tests pass on current numpy dev. Let's see what the CI says.

ping @seberg in case you want to look this over.


mpfdtype/mpfdtype/src/mpfdtype_main.c
```diff
@@ -47,10 +44,14 @@ PyInit__mpfdtype_main(void)
...
     }
 
-    if (init_mpf_umath() < 0) {
+    PyObject *numpy = init_mpf_umath();
+    
+    if (numpy == NULL) {
         goto error;
     }
 
+    Py_DECREF(numpy);
```
seberg, line 53: Not sure I understand the `numpy` return change here.

seberg (approved):
Thanks for looking into this!  Just scrolled through and looks fine.  I didn't understand why you added that `numpy` return, but I may have missed it elsewhere.

I guess for the string dtype deleting tests where NumPy changed makes sense (and is fine with me either way).

Please don't hesitate to just self merge when you got the tests passing.

mpfdtype/mpfdtype/src/mpfdtype_main.c
ngoldbaum: This was because I wasn't aware that there is a unique symbol macro for the ufunc API. Now I'm using that too and putting the import in the module init.

PR #98, opened 2024-08-07
Adding Quaddtype

SwayamInSync:
This PR adds the initial support of SLEEF backed QuadDtype


mattip:
Heh. Over at openblas-lib we were just discussing what would happen if someone started using the libquadmath functions linked from gfortran in https://github.com/MacPython/openblas-libs/pull/85. Will this eventually percolate into linalg calls? It would require a new build of OpenBLAS, since by default [support is disabled](https://github.com/OpenMathLib/OpenBLAS/blob/d92cc96978c17a35355101a1901981970dec25b6/docs/build_system.md?plain=1#L81) with an admonition that the feature is not well implemented.

How will this interact with numpy ufuncs, are you planning to add required loops?

shibatch:
Hello,

If you guys really care about correctly-rounded results, please take a look at tlfloat.

https://github.com/shibatch/tlfloat

It also has support for IEEE octuple-precision FP type.

shibatch:
tlfloat is a library positioned completely differently from sleef.
In a word, it is similar to mpfr, but tlfloat works everywhere.

ngoldbaum:
@mattip if we upstream this to NumPy (still lots of work before we can do that) we'd probably need to use SLEEF exclusively, which has the boost (BSD-like) license.

The plan is to add ufuncs and casts in followups.

If you'd like we should chat about this project sometime so you can meet Swayam.

ngoldbaum:
@shibatch thanks for your interest here! We'd love your opinion bout this work. 

The goal here is to maybe eventually replace numpy's longdouble dtype with something that doesn't consistently produce annoying platform-specific bugs that maintainers who do not care about longdouble need to fix. Also providing true 128 bit floats. NumPy has long had a float128 DType that is not really a float128.

tlfloat seems like a good option for the MPFR DType @seberg wrote, if anyone wants to push that farther forward.

ngoldbaum:
It's not correctly rounded results that we care about so much as a migration path for users of longdouble. Also a way to get higher precision floats from a cross-platform library.

Agreed about the speed of SLEEF. Is there something faster we can use? That said, speed is less of a concern IMO than providing the functionality at all. Right now in NumPy if you're not on Linux you're out of luck if you need anything higher precision than 64 bit floats.

We also talked about writing the code in such a way that you could select the float backend at runtime with a parameter for the dtype constructor, since that would help with benchmarking and experimentation.

ngoldbaum:
Oh, I misinterpreted your earlier message, tlfloat *does* implement quads.

shibatch:
@ngoldbaum Thank you for your interest in sleef and tlfloat.

I have not yet benchmarked the computation speed of tlfloat, but I believe it is similar to libquadmath in terms of the basic four arithmetic operations, since the internal processing is equivalent to libquadmath.

For elementary functions, sleef should be significantly faster. Sleef internally performs operations in triple-double format, and the reason it is slower for the basic four arithmetic operations is due to the slow conversion between triple-double and IEEE format.
For elementary functions, it performs many calculations with triple-doubles, so it is easier to gain a speed advantage.

quaddtype/tests/test_arr.py
ngoldbaum, line 1: I haven't looked carefully at the rest of this, but a quick comment is that I think we should still call this file `test_quaddtype`. Also didn't you mention that creating arrays is working? Probably worth adding a test that does that.

SwayamInSync: Yes, the current one also makes an array with `QuadPrecDType`, will push the separate in next commits

quaddtype/quaddtype/src/umath.cpp
```diff
@@ -0,0 +1,356 @@
...
+    return NPY_NO_CASTING;  // Quad precision is always the same precision
+}
+
+// todo: skipping the promoter for now, since same type operation will be requried
+
+template <binop_def binop>
+int
+create_quad_binary_ufunc(PyObject *numpy, const char *ufunc_name)
+{
+    PyObject *ufunc = PyObject_GetAttrString(numpy, ufunc_name);
```
ngoldbaum, line 197: you need to decref this in the error cases and before you return

quaddtype/quaddtype/src/umath.cpp
```diff
@@ -0,0 +1,356 @@
...
+    }
+    if (create_quad_binary_ufunc<quad_mod>(numpy, "mod") < 0) {
+        return -1;
+    }
+    return 0;
+}
+
+// comparison functions
+
+template <cmp_def comp>
```
ngoldbaum, line 252: similar to above, unnecessary template

quaddtype/quaddtype/src/umath.cpp
```diff
@@ -0,0 +1,356 @@
...
+        *((npy_bool *)out_ptr) = comp((Sleef_quad *)in1_ptr, (Sleef_quad *)in2_ptr);
+
+        in1_ptr += in1_stride;
+        in2_ptr += in2_stride;
+        out_ptr += out_stride;
+    }
+    return 0;
+}
+
+template <cmp_def comp>
```
ngoldbaum, line 275: here too

quaddtype/tests/test_quaddtype.py
```diff
@@ -1,33 +1,108 @@
...
+    op_func = getattr(operator, op)
+    quad_a = QuadPrecision("12.5")
+    quad_b = QuadPrecision(other)
+    float_a = 12.5
+    float_b = float(other)
+
+    quad_result = op_func(quad_a, quad_b)
+    float_result = op_func(float_a, float_b)
 
+    assert np.abs(np.float64(quad_result) - float_result) < 1e-10
```
ngoldbaum, line 38: I don't think this is testing much, let's chat more about how to test these ufuncs at our next meeting.

ngoldbaum (commented):
I left a bunch of comments inline but it's awesome to see how much progress you've already made!

quaddtype/quaddtype/src/dtype.c
```diff
@@ -1,197 +1,214 @@
...
+#define PY_UFUNC_UNIQUE_SYMBOL QuadPrecType_UFUNC_API
 #define NPY_NO_DEPRECATED_API NPY_2_0_API_VERSION
 #define NPY_TARGET_VERSION NPY_2_0_API_VERSION
 #define NO_IMPORT_ARRAY
 #define NO_IMPORT_UFUNC
-#include "numpy/ndarraytypes.h"
 #include "numpy/arrayobject.h"
+#include "numpy/ndarraytypes.h"
 #include "numpy/dtype_api.h"
+#include "numpy/_public_dtype_api_table.h" // not included in dtype_api.h
```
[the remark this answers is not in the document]
SwayamInSync: `PyArrayDTypeMeta_Type` is defined inside the `_public_dtype_api_table.h` for setting the `ob_type` as
```
    ((PyObject *)&QuadPrecDType)->ob_type = &PyArrayDTypeMeta_Type;
```

quaddtype/quaddtype/src/scalar_ops.cpp
```diff
@@ -0,0 +1,171 @@
...
+PyNumberMethods quad_as_scalar = {
+        .nb_add = (binaryfunc)quad_binary_func<quad_add>,
+        .nb_subtract = (binaryfunc)quad_binary_func<quad_sub>,
+        .nb_multiply = (binaryfunc)quad_binary_func<quad_mul>,
+        .nb_true_divide = (binaryfunc)quad_binary_func<quad_div>,
+        .nb_power = (ternaryfunc)quad_binary_func<quad_pow>,
+        .nb_negative = (unaryfunc)quad_unary_func<quad_negative>,
+        .nb_positive = (unaryfunc)quad_unary_func<quad_positive>,
+        .nb_absolute = (unaryfunc)quad_unary_func<quad_absolute>,
+        .nb_bool = (inquiry)quad_nonzero,
```
ngoldbaum, line 170: The error in CI is coming from the order in which these are declared. Under gcc you need to iniitialize c++ struct fields in the order they're declared in [the header](https://docs.python.org/3/c-api/typeobj.html#c.PyNumberMethods). The error is about `nb_power` because it comes before `nb_true_divide` in the header.

quaddtype/quaddtype/src/dtype.c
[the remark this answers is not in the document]
ngoldbaum: Ah, you should be doing `#include "numpy/ndarraytypes.h"` instead of including the private header. It's a core data type of NumPy so it can't live in the dtype API header.

quaddtype/tests/test_quaddtype.py
ngoldbaum: We didn't get a chance to talk about this last night. The idea I had was to use sympy and/or mpfloat to calculate these results with *arbitrary* precision, then make sure the result you get is within the expected error bounds (1-4 ULP depending on the operation IIRC in the SLEEF docs).

quaddtype/quaddtype/src/casts.cpp
```diff
@@ -0,0 +1,449 @@
...
+inline Sleef_quad
+to_quad<double>(double x)
+{
+    return Sleef_cast_from_doubleq1(x);
+}
+template <>
+inline Sleef_quad
+to_quad<long double>(long double x)
+{
+    return Sleef_cast_from_doubleq1(x);
```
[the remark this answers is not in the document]
SwayamInSync: I am not sure here, but as handling earlier we can use `memcpy` to copy the `long double` value to an aligned buffer before passing it to the `Sleef_cast_from_doubleq1`
that way it can handle the non-alignment, something like this
```c++
alignas(long double) long double aligned_x;
memcpy(&aligned_x, &x, sizeof(long double));
return Sleef_cast_from_doubleq1(aligned_x);
```

quaddtype/quaddtype/src/casts.cpp
```diff
@@ -0,0 +1,449 @@
...
+        }
+    }
+    else {
+        Py_INCREF(given_descrs[1]);
+        loop_descrs[1] = given_descrs[1];
+    }
+
+    loop_descrs[0] = PyArray_GetDefaultDescr(dtypes[0]);
+    *view_offset = 0;
+    return NPY_SAFE_CASTING;
```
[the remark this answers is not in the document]
SwayamInSync: Yeah that make sense looking on the sizes as,
A 64-bit integer has a maximum value of $2^{64} - 1$ for unsigned integers and $ -2^{63} $ to $ 2^{63} - 1 $ for signed integers. The `float128` format has a 113-bit significand (mantissa) so technically it should represent without rounding

quaddtype/quaddtype/src/casts.cpp
```diff
@@ -0,0 +1,449 @@
...
+quad_to_numpy_resolve_descriptors(PyObject *NPY_UNUSED(self), PyArray_DTypeMeta *dtypes[2],
+                                  PyArray_Descr *given_descrs[2], PyArray_Descr *loop_descrs[2],
+                                  npy_intp *view_offset)
+{
+    Py_INCREF(given_descrs[0]);
+    loop_descrs[0] = given_descrs[0];
+
+    loop_descrs[1] = PyArray_GetDefaultDescr(dtypes[1]);
+    *view_offset = 0;
+    return NPY_SAME_KIND_CASTING;
```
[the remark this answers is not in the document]
SwayamInSync: Getting it, Quad to other dtypes will lose the precision, changing it to `NPY_UNSAFE_CASTING`

quaddtype/quaddtype/src/casts.cpp
```diff
@@ -0,0 +1,449 @@
...
+        T *out_val = (T *)out_ptr;
+        *out_val = from_quad<T>(in_val);
+
+        in_ptr += strides[0];
+        out_ptr += strides[1];
+    }
+    return 0;
+}
+
+static std::vector<PyArrayMethod_Spec *> specs;
```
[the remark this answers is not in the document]
SwayamInSync: Right currently doing 29 (14 from, 14 to + 1 NULL) 
In the `free_casts` function, doing the `specs.clear()` can empty the vector, but will update this to C array as vector can eat up the unnecessary heap

ngoldbaum:
OK, tests are passing and I think all my comments have been addressed, let's bring this in. Let's try to keep followups more bite-sized to make reviewing these easier.

PR #100, opened 2024-08-23
Packaging QuadDType

SwayamInSync:
This PR adds the workflow for building wheels and making release for `quaddtype`

### DIFF
--- a/quaddtype/meson.build
+++ b/quaddtype/meson.build
@@ -5,8 +5,18 @@ py = py_mod.find_installation()
 
 c = meson.get_compiler('c')
 
-sleef_dep = c.find_library('sleef')
-sleefquad_dep = c.find_library('sleefquad')
+# Get the conda prefix
+conda_prefix = run_command('bash', '-c', 'echo $CONDA_PREFIX', check: true).stdout().strip()
+
+# Add conda lib directory to library path
+add_project_link_arguments('-L' + conda_prefix + '/lib', language: ['c', 'cpp'])
+
+sleef_dep = c.find_library('sleef', dirs: [conda_prefix + '/lib'])
+sleefquad_dep = c.find_library('sleefquad', dirs: [conda_prefix + '/lib'])
+
+if not sleef_dep.found() or not sleefquad_dep.found()
+    error('SLEEF library not found. Please ensure it is installed in your conda environment\nconda install sleef.')
+endif
 
 incdir_numpy = run_command(py,
   [
--- a/quaddtype/pyproject.toml
+++ b/quaddtype/pyproject.toml
@@ -13,7 +13,7 @@ name = "quaddtype"
 description = "Quad (128-bit) float dtype for numpy"
 version = "0.0.1"
 readme = 'README.md'
-author = "Swayam Singh"
+authors = [{name = "Swayam Singh", email = "singhswayam008@gmail.com"}]
 requires-python = ">=3.9.0"
 dependencies = [
     "numpy"
@@ -23,3 +23,40 @@ dependencies = [
 test = [
     "pytest",
 ]
+
+[tool.cibuildwheel]
+archs = ["auto64"]
+skip = ["*-musllinux*", "pp*", "cp36-*", "cp37-*", "cp38-*"]
+environment = { PATH = "$PATH:/usr/share/miniconda3/bin" }
+
+[tool.cibuildwheel.linux]
+before-all = """
+    wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
+    bash miniconda.sh -b -p /usr/share/miniconda3
+    export PATH="/usr/share/miniconda3/bin:$PATH"
+    conda config --add channels conda-forge
+    conda config --set channel_priority strict
+    conda install -y sleef
+"""
+
+[tool.cibuildwheel.macos]
+before-all = """
+    wget https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -O miniconda.sh
+    bash miniconda.sh -b -p $HOME/miniconda3
+    export PATH="$HOME/miniconda3/bin:$PATH"
+    conda config --add channels conda-forge
+    conda config --set channel_priority strict
+    conda install -y sleef
+"""
+
+[tool.cibuildwheel.windows]
+before-all = """
+    powershell -Command "(New-Object Net.WebClient).DownloadFile('https://repo.anaconda.com/miniconda/Miniconda3-latest-Windows-x86_64.exe', 'miniconda.exe')"
+    start /wait "" miniconda.exe /S /D=C:\\Miniconda3
+    set PATH=C:\\Miniconda3;C:\\Miniconda3\\Scripts;%PATH%
+    conda config --add channels conda-forge
+    conda config --set channel_priority strict
+    conda install -y sleef
+"""
+before-build = "pip install delvewheel"
+repair-wheel-command = "delvewheel repair -w {dest_dir} {wheel}"
--- a/.github/workflows/build_wheels_and_release.yml
+++ b/.github/workflows/build_wheels_and_release.yml
@@ -1,0 +1,98 @@
+name: Build quaddtype wheels and release
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'v*'
+  pull_request:
+
+jobs:
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, windows-2019, macos-11]
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version:  '3.10'
+
+      - name: Set up Conda
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          auto-update-conda: true
+          python-version: '3.10'
+          channels: conda-forge,defaults
+          channel-priority: strict
+
+      - name: Install SLEEF and other dependencies
+        shell: bash -l {0}
+        run: |
+          conda install -y sleef numpy meson meson-python patchelf wheel
+
+      - name: Install cibuildwheel
+        run: python -m pip install cibuildwheel==2.11.2
+
+      - name: Build wheels
+        shell: bash -l {0}
+        run: |
+          export LIBRARY_PATH=$CONDA_PREFIX/lib:$LIBRARY_PATH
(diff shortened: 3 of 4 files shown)


quaddtype/meson.build
```diff
@@ -1,20 +1,72 @@
...
+
+  sleef_include_dir = sleef_path + '/include'
+  sleef_library_dir = sleef_path + '/lib'
+
+  add_project_arguments('-I' + sleef_include_dir, language: ['c', 'cpp'])
+  add_project_link_arguments('-L' + sleef_library_dir, language: ['c', 'cpp'])
+  
+  sleef_dep = c.find_library('sleef', dirs: [sleef_library_dir])
+  sleefquad_dep = c.find_library('sleefquad', dirs: [sleef_library_dir])
+endif
```
ngoldbaum, line 50: I'm pretty sure the correct way to do this is to use `find_library` as you had it before. Why do you need all of this OS-specific stuff? Is it to find the conda package?

I think NumPy uses pkgconfig to avoid needing to spell out exactly how libraries were installed to build against them. You can use pkgconfig on windows in CI by installing pkgconfiglite. I've never actually needed to figure this out though so unfortunately can't help much. Ralf or someone else at Quansight probably knows a lot more about the "correct" way to handle third party deps in a meson.build on windows.

quaddtype/meson.build
```diff
@@ -1,20 +1,72 @@
...
+
+if not sleef_dep.found() or (not is_windows and not sleefquad_dep.found())
```
ngoldbaum, line 52: It seems wrong to not have an explicit `sleefquad_dep` on windows and adds unnecessary complication. What if someone tries to build against a verison of SLEEF without the quad support?

quaddtype/meson.build
```diff
@@ -47,10 +99,13 @@ py.install_sources(
...
 py.extension_module('_quaddtype_main',
-srcs,
-c_args: ['-g', '-O0', '-lsleef', '-lsleefquad'],
-dependencies: [sleef_dep, sleefquad_dep],
-install: true,
-subdir: 'quaddtype',
-include_directories: includes
+  srcs,
+  c_args: is_windows ? ['/DWIN32', '/D_WINDOWS'] : ['-g', '-O0'],
+  cpp_args: is_windows ? ['/DWIN32', '/D_WINDOWS', '/EHsc'] : ['-g', '-O0'],
```
ngoldbaum, line 104: You should pass `--buildtype=debug` to meson instead of hard-coding these.

quaddtype/pyproject.toml
```diff
@@ -23,3 +22,61 @@ dependencies = [
...
+# conda install -y -c conda-forge sleef
+# if [ ! -f "$HOME/miniconda/include/sleef.h" ]; then
+#     echo "sleef.h not found. Installation may have failed."
+#     exit 1
+# fi
+# ls -l $HOME/miniconda/include/sleef.h
+# ls -l $HOME/miniconda/lib/libsleef*
+# '''
+# environment = {DYLD_LIBRARY_PATH = "$HOME/miniconda/lib:$DYLD_LIBRARY_PATH", LIBRARY_PATH = "$HOME/miniconda/lib:$LIBRARY_PATH", CFLAGS = "-I$HOME/miniconda/include $CFLAGS", CXXFLAGS = "-I$HOME/miniconda/include $CXXFLAGS", LDFLAGS = "-L$HOME/miniconda/lib $LDFLAGS", SLEEF_INCLUDE_DIR = "$HOME/miniconda/include", SLEEF_LIBRARY = "$HOME/miniconda/lib", MACOSX_DEPLOYMENT_TARGET = "10.13"}
+# repair-wheel-command = "delocate-wheel -w {dest_dir} -v {wheel}"
```
ngoldbaum, line 82: Why is all of this configuration commented out?

.github/workflows/build_wheel.yml
```diff
@@ -0,0 +1,281 @@
...
+            LD_LIBRARY_PATH="/root/miniconda/envs/build_env/lib:$LD_LIBRARY_PATH"
+            LIBRARY_PATH="/root/miniconda/envs/build_env/lib:$LIBRARY_PATH"
+            CFLAGS="-I/root/miniconda/envs/build_env/include $CFLAGS"
+            CXXFLAGS="-I/root/miniconda/envs/build_env/include $CXXFLAGS"
+            LDFLAGS="-L/root/miniconda/envs/build_env/lib $LDFLAGS"
+            SLEEF_PATH="/root/miniconda/envs/build_env"
+          CIBW_REPAIR_WHEEL_COMMAND: >
+            auditwheel repair -w {dest_dir} --plat manylinux_2_28_x86_64 {wheel}
+          CIBW_TEST_COMMAND: |
+            python -c "import os; print('SLEEF_PATH:', os.environ.get('SLEEF_PATH', 'Not set'))"
```
ngoldbaum, line 54: Why test if `SLEEF_PATH` is set? That's just a detail of how you set up the build. Also if you delete this you can make the test command the same for all three architectures.

.github/workflows/build_wheel.yml
```diff
@@ -0,0 +1,281 @@
...
+            python -c "import platform; print('Python version:', platform.python_version())"
+            python -c "import sys; print('sys.platform:', sys.platform)"
+            python -c "import quaddtype; print('quaddtype imported successfully')"
+            pip install {package}[test]
```
ngoldbaum, line 58: Why not just install the test dependencies when you install it the first time?

.github/workflows/build_wheel.yml
```diff
@@ -0,0 +1,281 @@
...
+        shell: pwsh
+        run: |
+          python -m cibuildwheel --output-dir wheelhouse
+          if (-not (Test-Path wheelhouse/*.whl)) { throw "Wheel was not created" }
+        working-directory: ./quaddtype
+
+      - uses: actions/upload-artifact@v3
+        with:
+          path: ./quaddtype/wheelhouse/*.whl
+          name: wheels-windows-${{ matrix.architecture }}
```
ngoldbaum, line 229: I noted the duplication about the test commands, but there's a lot of other duplication as well. Can you try to unify the logic a bit? I think in principle you could make the Mac and Linux builds identical, only conditionally setting a few environment variables differently. The Windows build needs it own separate build section, but I don't think the dependency setup, testing, or upload steps need to be repeated.

It's OK to have architecture-specific steps within a longer workflow by branching on the build platform in the build matrix using like e.g. `if: matrix.platform == 'win32'`.

.github/workflows/build_wheel.yml
```diff
@@ -0,0 +1,281 @@
+name: Build Wheels
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "v*"
```
ngoldbaum, line 8: It's just a convention, but since there are multiple dtypes in the repo, I think it's less confusing to only look for tags like `quaddtype-v*`.

.github/workflows/build_wheel.yml
```diff
@@ -0,0 +1,281 @@
...
+  pull_request:
```
ngoldbaum, line 9: I don't think this needs to trigger for every pull request or push to main. First, it should only trigger if the `quaddtype` folder gets touched.

But also I think it should only trigger wheel builds if there's a tag like [wheel build] in the commit message or a git tag with a compatible name gets pushed.

ngoldbaum: Also probably worth adding `workflow_dispatch`.

quaddtype/meson.build
```diff
@@ -1,20 +1,72 @@
-project('quaddtype', 'c', 'cpp', default_options : ['cpp_std=c++17', 'b_pie=true'])
+project('quaddtype', 'c', 'cpp', default_options : ['cpp_std=c++20', 'b_pie=true'])
```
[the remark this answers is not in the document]
SwayamInSync: Getting this error in windows build (MSVC)

quaddtype/pyproject.toml
SwayamInSync: I put the configuration from there to workflow itself, so not needed there anymore

ngoldbaum: OK in that case you should delete it. IMO you should never commit commented out code to a public repo, it's always going to be a source of confusion and bitrotted code.

quaddtype/meson.build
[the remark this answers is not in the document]
ngoldbaum: I wonder if that will be problematic on visual studio 2019, which is still supported aiui. If it's not an issue here on CI I think it's fine, definitely something to check about with someone who has more experience distributing wheels that depend on C++ code.

quaddtype/pyproject.toml
SwayamInSync: Understood ✅

.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,249 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          path: dist
+      - name: Publish to TestPyPI
+        uses: pypa/gh-action-pypi-publish@v1.9.0
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_API_TOKEN }}
+          repository-url: https://test.pypi.org/legacy/
+          packages-dir: dist/*
```
ngoldbaum, line 215: @mattip when you have time, can you coordinate with Swayam to add the configuration necessary to upload to pypi at https://pypi.org/project/numpy_quaddtype/? I think [these are the relevant docs](https://docs.pypi.org/trusted-publishers/adding-a-publisher/).

@SwayamInSync you should also probably manually upload a numpy_quaddtype release to pypi with a tarball containingnthe current state of the main branch. Mostly just to claim the name, but I think the project may have to exist already on pypi before you can set up trusted publishing.

ngoldbaum (commented):
All the build changes look much cleaner now, thanks for removing all the hardcoding. I'm glad you were able to get the wheel builds working as well.

Another followup task after this is to create a new feedstock for numpy_quaddtype on conda-forge. Since SLEEF is already packaged that should be straightforward I think. If you've never set up a conda package before I can point you at the relevant docs.

Once that's done and we have wheels up on PyPI I think we can make an announcement on the numpy mailing list asking for feedback.

SwayamInSync:
> All the build changes look much cleaner now, thanks for removing all the hardcoding. I'm glad you were able to get the wheel builds working as well.
> 
> Another followup task after this is to create a new feedstock for numpy_quaddtype on conda-forge. Since SLEEF is already packaged that should be straightforward I think. If you've never set up a conda package before I can point you at the relevant docs.
> 
> Once that's done and we have wheels up on PyPI I think we can make an announcement on the numpy mailing list asking for feedback.

Yeah make sense, I never set-up a conda package in past, so it'll be very helpful with the relevant docs :)

ngoldbaum:
I'll try to look at this again soon, sorry for letting it fall of my radar.

PR #102, opened 2024-09-16
Updating Quaddtype

SwayamInSync:
This PR updates the following to quaddtype
- Support for second `longdouble` backend
- more unary and binary funcs
- renamed to `numpy_quaddtype`
- Dragon4 printing

and more small refactors


quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -204,91 +320,162 @@ numpy_to_quad_strided_loop(PyArrayMethod_Context *context, char *const data[],
...
+from_quad<long double>(quad_value x, QuadBackendType backend)
 {
-    return Sleef_cast_to_doubleq1(x);
+    if (backend == BACKEND_SLEEF) {
+        return (long double)Sleef_cast_to_doubleq1(x.sleef_value);
+    }
+    else {
+        return x.longdouble_value;
+    }
 }
```
ngoldbaum, line 479: I think it would be simpler to template over backend and output type, then you wouldn't need the runtime branching.

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -315,10 +501,17 @@ quad_to_numpy_strided_loop(PyArrayMethod_Context *context, char *const data[],
...
+
     while (N--) {
-        Sleef_quad in_val = *(Sleef_quad *)in_ptr;
-        T *out_val = (T *)out_ptr;
-        *out_val = from_quad<T>(in_val);
+        quad_value in_val;
+        memcpy(&in_val, in_ptr, elem_size);
+
+        T out_val = from_quad<T>(in_val, backend);
+        memcpy(out_ptr, &out_val, sizeof(T));
```
ngoldbaum, line 514: future improvement would be to define both unaligned and aligned loops and only memcpy in the unaligned loops

quaddtype/numpy_quaddtype/src/dragon4.c
ngoldbaum, line 1: you should add a comment that this code was extracted from NumPy and the original author was @ahaldane (who may also be interested in this DType :))

quaddtype/numpy_quaddtype/src/dtype.c
```diff
@@ -0,0 +1,261 @@
...
+quad_store(char *data_ptr, void *x, QuadBackendType backend)
+{
+    if (data_ptr == NULL || x == NULL) {
+        return -1;
+    }
+    if (backend == BACKEND_SLEEF) {
+        *(Sleef_quad *)data_ptr = *(Sleef_quad *)x;
+    }
+    else {
+        *(long double *)data_ptr = *(long double *)x;
```
ngoldbaum, line 45: You can avoid the branching here and in `quad_load` if you split them into separate functions for longdouble and quad. I would expect these functions might get called a lot and repeatedly so avoiding branching should improve performance.

quaddtype/numpy_quaddtype/src/dtype.c
```diff
@@ -0,0 +1,261 @@
...
+    return self;
+}
+
+static QuadPrecDTypeObject *
+common_instance(QuadPrecDTypeObject *dtype1, QuadPrecDTypeObject *dtype2)
+{
+    if (dtype1->backend != dtype2->backend) {
+        PyErr_SetString(PyExc_TypeError,
+                        "Cannot find common instance for QuadPrecDTypes with different backends");
+        return NULL;
```
ngoldbaum, line 85: This is fine but if anyone mentions it in the future, I think it would probably also work to return a SLEEF quad, since it's always safe to cast a long double to a quad.

quaddtype/numpy_quaddtype/src/ops.hpp
```diff
@@ -0,0 +1,513 @@
...
+{
+    *out = (*in1 > *in2) ? *in1 : *in2;
+    return 0;
+}
+
+static inline int
+ld_atan2(long double *out, long double *in1, long double *in2)
+{
+    *out = atan2l(*in1, *in2);
+    return 0;
```
ngoldbaum, line 434: Is it possible for any of the functions from here up implementing various operations to ever return anything but 0? I don't see anything... In that case it probably makes more sense to just return the result of the operation and not have unncessary error codes.

quaddtype/numpy_quaddtype/src/scalar.c
```diff
@@ -0,0 +1,253 @@
...
+        if (backend == BACKEND_SLEEF) {
+            self->value.sleef_value = Sleef_cast_from_int64q1(val);
+        }
+        else {
+            self->value.longdouble_value = (long double)val;
+        }
+    }
+    else if (Py_TYPE(value) == &QuadPrecision_Type)
+    {
+        // todo: not working for ld backend, getting garbage value not sure why?
```
ngoldbaum, line 81: can you add a failing test for this to the test suite? I can try to poke around at fixing it...

quaddtype/numpy_quaddtype/src/umath.cpp
```diff
@@ -0,0 +1,815 @@
...
+            sleef_op((Sleef_quad *)in_ptr, (Sleef_quad *)out_ptr);
+        }
+        else {
+            longdouble_op((long double *)in_ptr, (long double *)out_ptr);
+        }
+        in_ptr += in_stride;
+        out_ptr += out_stride;
+    }
+    return 0;
+}
```
ngoldbaum, line 164: nice! i'm glad it was easy to implement both versions

quaddtype/numpy_quaddtype/src/umath.cpp
```diff
@@ -0,0 +1,815 @@
...
+    npy_intp in1_stride = strides[0];
+    npy_intp in2_stride = strides[1];
+    npy_intp out_stride = strides[2];
+
+    QuadPrecDTypeObject *descr = (QuadPrecDTypeObject *)context->descriptors[0];
+    QuadBackendType backend = descr->backend;
+
+    while (N--) {
+        if (backend == BACKEND_SLEEF) {
+            sleef_op((Sleef_quad *)out_ptr, (Sleef_quad *)in1_ptr, (Sleef_quad *)in2_ptr);
```
ngoldbaum, line 384: Here and in the unary op implementation, there's a chance that either or both `in1_ptr` and `in2_ptr` are the same location as `out_ptr`. Is that problematic? You should be able to test that by making sure that e.g. `np.multiply(a, b, out=a)` or `np.multiply(a, a, out=a)` is doing the right thing.

ngoldbaum (commented):
I did not carefully review all the reference counting or go over everything with a fine-toothed comb but I did go over the whole thing and followed along with the structure.

Overall this looks really nice. I think there are some clear paths to improve it, but that's always going to be the case and this is already far enough along that once we have some basic docs written we can try inviting people to test it and start thinking about a NEP to do something about np.128.

quaddtype/numpy_quaddtype/src/umath.cpp
SwayamInSync: Getting the expected results for both ops
```bash
In [14]: a = np.array([1,2,3], dtype=npq.QuadPrecDType())

In [15]: b = np.array([4,5,6], dtype=npq.QuadPrecDType())

In [16]: np.multiply(a, b, out=a)
Out[16]: 
array([QuadPrecision('4.0e+000', backend='sleef'),
       QuadPrecision('1.0e+001', backend='sleef'),
       QuadPrecision('1.8e+001', backend='sleef')],
      dtype=QuadPrecDType(backend='sleef'))

In [17]: a
Out[17]: 
array([QuadPrecision('4.0e+000', backend='sleef'),
       QuadPrecision('1.0e+001', backend='sleef'),
       QuadPrecision('1.8e+001', backend='sleef')],
      dtype=QuadPrecDType(backend='sleef'))

In [18]: np.multiply(a, a, out=a)
Out[18]: 
array([QuadPrecision('1.6e+001', backend='sleef'),
       QuadPrecision('1.0e+002', backend='sleef'),
       QuadPrecision('3.24e+002', backend='sleef')],
      dtype=QuadPrecDType(backend='sleef'))

In [19]: a
Out[19]: 
array([QuadPrecision('1.6e+001', backend='sleef'),
       QuadPrecision('1.0e+002', backend='sleef'),
       QuadPrecision('3.24e+002', backend='sleef')],
      dtype=QuadPrecDType(backend='sleef'))
```

quaddtype/numpy_quaddtype/src/casts.cpp
SwayamInSync: Ahh this must have been missed, looking into that

quaddtype/numpy_quaddtype/src/scalar.c
SwayamInSync: This is fixed now, it wasn't working because by default SLEEF backend Quad was creating for all inputs leading to garbage value.

Will push the future edits by today along with the review resolves

quaddtype/numpy_quaddtype/src/casts.cpp
SwayamInSync: This is resolved

quaddtype/numpy_quaddtype/src/ops.hpp
SwayamInSync: This is resolved

quaddtype/README.md
```diff
@@ -0,0 +1,22 @@
...
+import numpy as np
+from numpy_quaddtype import QuadPrecDType, QuadPrecision
+
+# using sleef backend (default)
+np.array([1,2,3], dtype=QuadPrecDType())
+np.array([1,2,3], dtype=QuadPrecDType("sleef"))
+
+# using longdouble backend
+np.array([1,2,3], dtype=QuadPrecDType("longdouble"))
+```
```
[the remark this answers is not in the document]
ngoldbaum: Oh that’s too bad. Before we do a release let’s also try to improve the tests.

PR #104, opened 2025-02-27
Add install from source instructions

mhvk:
I thought I would give the quad dtype a try. It was giving errors (will raise separate issue), so I thought I should perhaps install from source. That doesn't really help, but as it took me a while to figure out how to do it, it seemed a good idea to add instructions to the `README`. Note that this is super linux-centric, but perhaps better than nothing?

### DIFF
--- a/quaddtype/README.md
+++ b/quaddtype/README.md
@@ -20,3 +20,39 @@ np.array([1,2,3], dtype=QuadPrecDType("sleef"))
 # using longdouble backend
 np.array([1,2,3], dtype=QuadPrecDType("longdouble"))
 ```
+
+## Install from source
+
+The code needs the quad precision pieces of the sleef library, which
+is not available on most systems by default, so we have to generate
+that first.  The below assumes one has the required pieces to build
+sleef (cmake and libmpfr-dev), and that one is in the package
+directory locally.
+
+```
+git clone https://github.com/shibatch/sleef.git
+cd sleef
+cmake -S . -B build -DSLEEF_BUILD_QUAD:BOOL=ON -DSLEEF_BUILD_SHARED_LIBS:BOOL=ON -DCMAKE_POSITION_INDEPENDENT_CODE=ON
+cmake --build build/ --clean-first -j
+cd ..
+```
+
+In principle, one can now install this system-wide, but easier would
+seem to use the version that was just created, as follows:
+```
+export SLEEF_DIR=$PWD/sleef/build
+export LIBRARY_PATH=$SLEEF_DIR/lib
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Fun, not sure why the test is failing, but it's unrelated.

@SwayamInSync would you mind giving this a look? If not I'll try diving in here a bit tomorrow (and see if I can reproduce @mhvk's crash).

mhvk:
OK, sounds good. Note that I basically started from the github workflow; great that you had that!

SwayamInSync:
> Fun, not sure why the test is failing, but it's unrelated.
> 
> @SwayamInSync would you mind giving this a look? If not I'll try diving in here a bit tomorrow (and see if I can reproduce @mhvk's crash).

Sure, I'll give this a go this weekend, a possible hunch is that Sleef got some updates recently (didn't followed them thoroughly) but that can be a possibility of crash.
Thanks @mhvk we will check and fix it :)

mhvk:
@SwayamInSync - thanks!  As for sleef, after I installed sleef on my machine, I did run the tests and they all passed.

SwayamInSync:
### Quick fix and issue
I see, my suspection seems to be correct, it was indeed version changing issues (and a pretty interesting one)
We build this with the 3.6 version of Sleef and now currently ongoing in 3.8

So I changed the version back to 3.6 in my fork and it passed all the tests 
```bash
git clone -b 3.6 https://github.com/shibatch/sleef.git
```

### Now fun part:
The issue is in the that test too, the test checks for the `nan` as 
```python
assert (QuadPrecision("nan") != QuadPrecision("nan")) == (QuadPrecision("nan") == QuadPrecision("nan"))
```
A simplified version would be
```python
(nan != nan) == (nan == nan)
```

As per IEEE-754 this should evaluate to
```python
(nan != nan)  # True
(nan == nan)  # False

# so overall
True == False
```

So the previous version (3.6) was IEEE incorrect (passing that test) and which when fixed in 3.8 made the error

so the correct test should be
```python
def test_nan_and_inf():
    # NaN should not equal itself
    assert QuadPrecision("nan") != QuadPrecision("nan")
    
    # Test infinity comparisons
    assert QuadPrecision("inf") > QuadPrecision("1e1000")
    assert QuadPrecision("-inf") < QuadPrecision("-1e1000")
```

@ngoldbaum another reason we should plan a tlfloat backend :)

mhvk:
I now checked on my laptop and it turns out my installation instructions had some typos and had aftereffects of experimenting with environment variables. I corrected these.

ngoldbaum:
Thanks @mhvk, let's fix the failing CI in a separate PR.

PR #108, opened 2025-03-05
fix nan comparison test

ngoldbaum:
This should fix the test that's failing on CI


ngoldbaum:
StringDType had tests running address sanitizer which relied on `LD_PRELOAD` and that seems to no longer work.

I turned off the sanitizers. The NumPy tests do the same job with code that people actually use.

PR #114, opened 2025-07-03
NaN tests + Workflow + README fixes

SwayamInSync:
This PR contributes as follows:
- Sleef comparison operators with NaN return false on equal and not equal comparisons (unlike the proposition [NaN-wikipedia](https://en.wikipedia.org/wiki/NaN#Comparison_with_NaN) I earlier found in IEEE-754 that says `nan != nan` should return true) hence removing that particular test
- Updated the README file with latest instructions to follow for installation
- Modifies workflow to publish on PyPI rather than Test-PyPI

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -194,8 +194,8 @@ jobs:
           path: ./quaddtype/wheelhouse/*.whl
           name: wheels-windows-${{ matrix.architecture }}
 
-  publish_to_testpypi:
-    name: Publish to TestPyPI
+  publish_to_pypi:
+    name: Publish to PyPI
     needs: [build_wheels_linux, build_wheels_macos, build_wheels_windows]
     runs-on: ubuntu-latest
     if: startsWith(github.ref, 'refs/tags/quaddtype-v')
@@ -204,12 +204,11 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           path: dist
-      - name: Publish to TestPyPI
+      - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@v1.9.0
         with:
           user: __token__
           password: ${{ secrets.PYPI_API_TOKEN }}
-          repository-url: https://test.pypi.org/legacy/
           packages-dir: dist/*
 
   create_release:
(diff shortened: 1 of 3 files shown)


.github/workflows/build_wheels.yml
```diff
@@ -204,12 +204,11 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           path: dist
-      - name: Publish to TestPyPI
+      - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@v1.9.0
         with:
           user: __token__
           password: ${{ secrets.PYPI_API_TOKEN }}
```
ngoldbaum, line 211: It looks like the latest version of the docs say not to use this anymore: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

juntyr:
I've opened #115 to fix the NaN comparisons and add extra tests for them

quaddtype/README.md
```diff
@@ -21,38 +23,32 @@ np.array([1,2,3], dtype=QuadPrecDType("sleef"))
...
-seem to use the version that was just created, as follows:
-```
+Building the `numpy-quaddtype` package from locally installed sleef:
+```bash
 export SLEEF_DIR=$PWD/sleef/build
 export LIBRARY_PATH=$SLEEF_DIR/lib
 export C_INCLUDE_PATH=$SLEEF_DIR/include
 export CPLUS_INCLUDE_PATH=$SLEEF_DIR/include
-python3 -m venv temp
-source temp/bin/activate
```
ngoldbaum, line 48: I think you should keep this

quaddtype/README.md
```diff
@@ -21,38 +23,32 @@ np.array([1,2,3], dtype=QuadPrecDType("sleef"))
...
 pip install meson-python numpy pytest
 pip install -e . -v --no-build-isolation
 export LD_LIBRARY_PATH=$SLEEF_DIR/lib
 ```
 
-Here, we created an editable install on purpose, so one can just work
-from the package directory if needed, e.g., to run the tests with,
-```
-python -m pytest
-```
```
ngoldbaum, line 58: and this

ngoldbaum (commented):
I think two of the changes to the readme should be reverted

SwayamInSync:
Just noticed window's build time is increased to 10 minutes but this seems to work

ngoldbaum:
Thanks @SwayamInSync!

PR #115, opened 2025-07-04
Test and fix comparisons with NaNs

juntyr:
Inspired by #114


SwayamInSync:
From logs it seem building from cmake using `--parallel` causing this race condition. I'll debug this and push a fix

SwayamInSync:
@ngoldbaum windows CI build fix was pushed in #114

PR #118, opened 2025-07-20
Implement sign ufunc extend unary op tests

juntyr:


juntyr:
Ideally, I'd also like to add signbit, but I'm unsure about how to implement it for sleef

SwayamInSync:
> Ideally, I'd also like to add signbit, but I'm unsure about how to implement it for sleef

So there is a way to extract sign bit from sleef_quad value, I used that in `dragon4.c` here is the link to those lines 
https://github.com/numpy/numpy-user-dtypes/blob/591481101ccf09b35b3d9ce9b282b13a10d7c5c7/quaddtype/numpy_quaddtype/src/dragon4.c#L1867C5-L1874C18
So you can simplify it as follows and it should work
```cpp
static inline int32_t
quad_signbit(Sleef_quad *op)
{
    union
    {
        Sleef_quad q;
        struct
        {
            uint64_t lo;
            uint64_t hi;
        } i;
    } u;
    u.q = *op;

    // Sign bit is the MSB of the high 64 bits
    return (u.i.hi >> 63) != 0;
}
```

juntyr:
Is the ordering of hi and lo guaranteed independent of endianness?

SwayamInSync:
With that I'll also suggest when writing pytest tests, use string based inputs (casting from double sometimes give precision issues)

ngoldbaum:
We test little endian architectures in the numpy CI using PPC64le QEMU emulation. It's slow but it at least lets you do a real end-to-end test.

PR #123, opened 2025-07-26
Implement the fmin and fmax ufuncs

juntyr:
- Implements the `fmin` and `fmax` ufuncs
- Unblocks using `np.nanmin` and `np.nanmax`

### DIFF
--- a/quaddtype/numpy_quaddtype/src/ops.hpp
+++ b/quaddtype/numpy_quaddtype/src/ops.hpp
@@ -399,6 +399,22 @@ quad_maximum(const Sleef_quad *in1, const Sleef_quad *in2)
 }
 
 static inline Sleef_quad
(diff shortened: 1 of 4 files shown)


SwayamInSync:
Thanks @juntyr , I wonder registering `quad_minimum/maximum` with `fmin` `fmax` ufuncs would've also worked?

juntyr:
> Thanks @juntyr , I wonder registering `quad_minimum/maximum` with `fmin` `fmax` ufuncs would've also worked?

No, they have different NaN propagation behaviour. minimum/maximum propagates NaNs, fmin/fmax tries to avoid propagating them

SwayamInSync:
> No, they have different NaN propagation behaviour. minimum/maximum propagates NaNs, fmin/fmax tries to avoid propagating them

Ah yes I just noticed, thanks merging it

Issue #113, opened 2025-07-03
SleefQuadPrecision np.mod ufunc gives unexpected results


juntyr:
The numpy docs state about np.mod that

```
It is equivalent to the Python modulus operator x1 % x2 and has the same sign as the divisor x2
```

For [-1.1, -1.0, -0.9, -0.1, 0.0, 0.1, 0.9, 1.0, 1.1], float64 mod +1 gives [0.9, 0.0, 0.1, 0.9, 0.0, 0.1, 0.9, 0.0, 0.1] and mod -1 [-0.1, -0.0, -0.9, -0.1, -0.0, -0.9, -0.1, -0.0, -0.9], but sleef gives [-0.1, -0.0, -0.9, -0.1, 0.0, 0.1, 0.9, 0.1] for both mod +1 and mod -1.

juntyr:
Fixed by #125

PR #125, opened 2025-08-03
Fixing mod's behaviour to numpy from IEEE

SwayamInSync:
This PR contributes as follows:
- Fixes the #113 
- Adds `np.remainder` (which is just alias of np.mod so no need to re-register)
- test suite for mod


juntyr (approved):

SwayamInSync:
nevermind, switched to qutil's NAN representation

PR #128, opened 2025-08-08
Implement hyperbolic ufuncs

juntyr:
- Implements the `sinh`, `cosh`, `tanh`, `arcsinh`, `arccosh`, and `arctanh` ufuncs
- Adds tests for the hyperbolic ufuncs
- Adds extra tests for handling `-0.0`
- Guarantee zero sign sensitive `minimum`, `maximum`, `fmin`, `fmax`

### DIFF
--- a/quaddtype/numpy_quaddtype/src/ops.hpp
+++ b/quaddtype/numpy_quaddtype/src/ops.hpp
@@ -151,6 +151,42 @@ quad_atan(const Sleef_quad *op)
     return Sleef_atanq1_u10(*op);
 }
 
+static inline Sleef_quad
+quad_sinh(const Sleef_quad *op)
+{
+    return Sleef_sinhq1_u10(*op);
+}
+
+static inline Sleef_quad
+quad_cosh(const Sleef_quad *op)
+{
+    return Sleef_coshq1_u10(*op);
+}
+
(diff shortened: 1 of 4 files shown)


juntyr:
@SwayamInSync Once we've figured out a way to handle the trigonometric ufuncs correctly, I'd like to add the hyperbolic ones as well - but I'll wait with writing tests until we have the trigonometric ones figured out

SwayamInSync:
Got it, thanks @juntyr

juntyr:
@SwayamInSync So sinh, cosh, tanh, and arctanh handle the large inputs as our tests expect. But for arcsinh and arccosh the floating result can overflow while the quad result doesn't yet. How would you suggest that we check that?

SwayamInSync:
> @SwayamInSync So sinh, cosh, tanh, and arctanh handle the large inputs as our tests expect. But for arcsinh and arccosh the floating result can overflow while the quad result doesn't yet. How would you suggest that we check that?

1e1000 and + are beyound float limits, we are good if they are working till one big common range i.e. `1e100` you can only test till this and we can take this in

juntyr:
@SwayamInSync From my end this should be ready to be merged now :)

SwayamInSync (approved):

PR #132, opened 2025-08-20
Add bits, precision, resolution constants

juntyr:
The precision is computed the same as numpy does for finfo, though I'm a bit unsure about the numpy import

### DIFF
--- a/quaddtype/numpy_quaddtype/__init__.py
+++ b/quaddtype/numpy_quaddtype/__init__.py
@@ -13,10 +13,12 @@ __all__ = [
     'SleefQuadPrecDType', 'LongDoubleQuadPrecDType', 'is_longdouble_128', 
     # Constants
     'pi', 'e', 'log2e', 'log10e', 'ln2', 'ln10', 'max_value', 'epsilon',
-    'smallest_normal', 'smallest_subnormal',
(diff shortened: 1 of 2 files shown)


juntyr:
?r @SwayamInSync

quaddtype/numpy_quaddtype/__init__.py
```diff
@@ -40,3 +42,6 @@ def LongDoubleQuadPrecDType():
 epsilon = get_sleef_constant("epsilon")
 smallest_normal = get_sleef_constant("smallest_normal")
 smallest_subnormal = get_sleef_constant("smallest_subnormal")
+bits = SleefQuadPrecDType().itemsize * 8
+precision = int(-np.log10(epsilon))
```
SwayamInSync, line 46: since `epsilon` is the sleef constant, you can use the `quad_log10` function instead of numpy and then just expose these in python and fetch them  here via `get_sleef_constant`.
That way if somewhere it'll be needed internally we can just utilize that avoiding recalculation.

SwayamInSync: Maybe can do with other constants as well, and I think it'll be a good idea to keep all the constants inside the main source code right?

Issue #130, opened 2025-08-19
No sdist on pypi


ngoldbaum:
Currently there's no sdist available on pypi, so `pip install numpy-quaddtype` fails for Python 3.13 because there aren't Python 3.13 wheels either.

We should set up wheels for 3.13 and 3.14 and make sure the sdist gets uploaded by the release automation.

juntyr:
I'd be very happy if we actually released a new version now. While it doesn't contain everything that I want yet, it would be good to test all the things that we've worked on irl

ngoldbaum:
Unfortunately I'm pretty swamped right now so I can't commit time to work on this for at least a week or two.

If you have time and interest then I should have time to at least review fixes for the release automation. It'd also be nice to add Python 3.13 and 3.14 builds too. I don't know what the SLEEF thread safety story is, but maybe free-threaded builds as well?

ngoldbaum:
> On a dev branch I can expand the Python in CI and test the builds for 3.13 and 3.14, given this I am also occupied in my "job's" work so this upcoming week might be tight for me to work.

Awesome. Obviously your day job takes precedence. I really appreciate the time you put into this. I may also try to poke at the release automation if I have some spare cycles in the next few weeks. There's no time pressure on this.

ngoldbaum:
Fixed by #135

PR #135, opened 2025-08-27
Building and uploading sdist along with prebuilt wheels

SwayamInSync:
Continuing #130

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -247,9 +247,71 @@ jobs:
           path: ./quaddtype/wheelhouse/*.whl
           name: wheels-windows-${{ matrix.architecture }}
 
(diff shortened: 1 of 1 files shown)


SwayamInSync:
@ngoldbaum continuing from #134

SwayamInSync:
> I think it should be possible to build the sdist without compiling sleef or the dtype but that seems to be how it's wired up now, so this is probably fine. There's probably a way to set up the meson config to fix that. But also I doubt anyone will care because the sdist is getting generated by a CI job.
> 
> Anyway, all that to say, looks good! Please go ahead and tag a release :)

Yes I tried doing without SLEEF and got the following meson error
```
../meson.build:17:4: ERROR: C shared or static library 'sleef' not found
```
 will check out if it's possible via some config

PR #143, opened 2025-09-05
Vendoring submodules using meson wrap [sdist compatible]

SwayamInSync:
This PR is my 1st attempt to vendor submodules dependencies within the repo as well as with sdist.

This also simplifies the GitHub workflows @ngoldbaum please take a look, if setup looks good I'll try to do same with sleef

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -137,3 +137,7 @@ compile_commands.json
 *.whl
 .DS_Store
 .idea/
+
+# quddtype
+/quaddtype/subprojects/qblas/
+.wraplock
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -21,18 +21,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-        with:
-          submodules: recursive
 
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
           python-version: ">=3.11.0"
-
-      - name: Verify QuadBLAS submodule
-        run: |
-          ls -la quaddtype/numpy_quaddtype/QBLAS/
-          ls -la quaddtype/numpy_quaddtype/QBLAS/include/quadblas/
 
       - name: Install cibuildwheel
         run: pip install cibuildwheel==3.1.4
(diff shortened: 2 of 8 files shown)


SwayamInSync:
My one wild guess is that sleef available on conds-forge is causing some issues with our tests so vendoring ours within sdist and building everything from there maybe solves them free

JaRoSchm:
Thank you for trying this! If someone else wants to have a look at the failure of one test using sleef from conda forge, please see https://github.com/conda-forge/numpy_quaddtype-feedstock/pull/4
We are happy about every suggestion helping us to fix this.

SwayamInSync:
> The README needs updates too, right? It would help me review if you update that too.

Sure, doing it.
Right now it is just `python -m pip install . -v`

.gitignore
```diff
@@ -137,3 +137,8 @@ compile_commands.json
 *.whl
 .DS_Store
 .idea/
+
+# quddtype
```
juntyr, line 141: quaddtype

quaddtype/README.md
```diff
@@ -25,28 +25,18 @@ np.array([1,2,3], dtype=QuadPrecDType("longdouble"))
...
-```bash
-git clone --branch 3.8 https://github.com/shibatch/sleef.git
-cd sleef
-cmake -S . -B build -DSLEEF_BUILD_QUAD:BOOL=ON -DSLEEF_BUILD_SHARED_LIBS:BOOL=ON -DCMAKE_POSITION_INDEPENDENT_CODE=ON
-cmake --build build/ --clean-first -j
-cd ..
-```
+### Linux/Unix/macOS
 
 Building the `numpy-quaddtype` package from locally installed sleef:
```
juntyr, line 37: Do we always assume that sleef is already locally installed? If not, now does the command differ if it isn't?

quaddtype/sdist_test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -x
+
+if [ -d "build/" ]; then
+    rm -rf dist/
+fi
+
+python -m pip uninstall -y numpy_quaddtype
+python -m build --sdist --outdir dist/
+python -m pip install dist/numpy_quaddtype-0.1.0.tar.gz -v
```
juntyr, line 10: Could we avoid hard coding the version into the path?

SwayamInSync:
These new commits just add an additional testing on top of the built sdist, to install and run pytest

quaddtype/sdist_test.sh
SwayamInSync: Reason behind that, you can't test both local build and sdist together as for changes to be visible in sdist we need to commit those changes first only then they get reflected and ready to test inside sdist. So I thought to do it separately.

juntyr (commented):
I cannot comment on the packaging and build details, but from a glance things look good to me

SwayamInSync:
@ngoldbaum can you take a look at this too please

juntyr:
> @ngoldbaum after test passes, you can merge this and we can do another release so that me and @JaRoSchm can use these upstream changes to test on conda-forge

Another release sounds good to me, perhaps we can wait until #145 is merged?

PR #145, opened 2025-09-07
Implement cast support for ubyte and half

juntyr:
Fixes numpy/numpy-quaddtype#27

### DIFF
--- a/quaddtype/tests/test_quaddtype.py
+++ b/quaddtype/tests/test_quaddtype.py
@@ -65,13 +65,11 @@ def test_supported_astype(dtype):
 
 @pytest.mark.parametrize("dtype", ["S10", "U10", "T", "V10", "datetime64[ms]", "timedelta64[ms]"])
 def test_unsupported_astype(dtype):
-    val = 1 if dtype != "V10" else b"1"
-
-    with pytest.raises(TypeError if dtype != "V10" else ValueError, match="cast"):
-        np.array(val, dtype=dtype).astype(QuadPrecDType, casting="unsafe")
-
     if dtype == "V10":
-        pytest.xfail("cast from QuadPrecision to V10 segfaults")
+        pytest.xfail("casts to and from V10 segfault")
+
+    with pytest.raises(TypeError, match="cast"):
+        np.array(1, dtype=dtype).astype(QuadPrecDType, casting="unsafe")
 
     with pytest.raises(TypeError, match="cast"):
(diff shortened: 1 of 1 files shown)


juntyr:
The other problem, and probably why ubyte wasn't added earlier, is that ubyte and bool alias but need different code paths. Not sure how to fix that ...

juntyr:
@SwayamInSync @ngoldbaum I think everything should work now, what do you think?

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -150,15 +151,26 @@ quad_to_quad_strided_loop_aligned(PyArrayMethod_Context *context, char *const da
...
 
 template <typename T>
 static inline quad_value
-to_quad(T x, QuadBackendType backend);
+to_quad(typename NpyType<T>::TYPE x, QuadBackendType backend);
 
 template <>
 inline quad_value
-to_quad<npy_bool>(npy_bool x, QuadBackendType backend)
+to_quad<my_npy_bool>(npy_bool x, QuadBackendType backend)
```
SwayamInSync, line 173: maybe rename these `my_`* template variable to something, rest looks fine

SwayamInSync: maybe starting with `spec_` (i.e. specialized)

quaddtype/tests/test_quaddtype.py
```diff
@@ -39,6 +39,42 @@ def test_finfo_int_constant(name, value):
...
+    back = quad.astype(dtype, casting="unsafe")
+
+    assert quad == 1
+    assert back == orig
+
+
+@pytest.mark.parametrize("dtype", ["S10", "U10", "T", "V10", "datetime64[ms]", "timedelta64[ms]"])
+def test_unsupported_astype(dtype):
+    if dtype == "V10":
+        pytest.xfail("casts to and from V10 segfault")
```
juntyr, line 69: I have no idea why this cast would segfault @ngoldbaum

ngoldbaum: Ah, you skipped it, never mind.

OK, this all looks good to me!

ngoldbaum (approved):

SwayamInSync:
BTW if needed then in this https://github.com/SwayamInSync/numpy-user-dtypes/pull/13 I wrote the workaround to get an "Non-Implemented" error instead of segfault crash.

PR #147, opened 2025-09-09
Make void to quad give error than segfault

SwayamInSync:
Follow up of #145

### DIFF
--- a/quaddtype/numpy_quaddtype/src/casts.cpp
+++ b/quaddtype/numpy_quaddtype/src/casts.cpp
@@ -21,7 +21,7 @@ extern "C" {
 #include "casts.h"
 #include "dtype.h"
 
-#define NUM_CASTS 33  // 16 to_casts + 16 from_casts + 1 quad_to_quad
+#define NUM_CASTS 34  // 16 to_casts + 16 from_casts + 1 quad_to_quad + 1 void_to_quad
 
 static NPY_CASTING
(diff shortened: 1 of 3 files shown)


SwayamInSync:
Let's do a release after this, what should be the versioning name?

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -151,6 +151,27 @@ quad_to_quad_strided_loop_aligned(PyArrayMethod_Context *context, char *const da
...
+    return (NPY_CASTING)-1;
+}
+
+static int
+void_to_quad_strided_loop(PyArrayMethod_Context *context, char *const data[],
+                         npy_intp const dimensions[], npy_intp const strides[],
+                         void *NPY_UNUSED(auxdata))
+{
+    PyErr_SetString(PyExc_RuntimeError, "void_to_quad_strided_loop should not be called");
+    return -1;
```
ngoldbaum, line 171: We could probably also do something similar in NumPy, but this is fine for now.

ngoldbaum:
Probably 0.2 because of all the packaging changes.

SwayamInSync:
> Probably 0.2 because of all the packaging changes.

Got it

SwayamInSync:
GitHub is depreciating the macos-13 runner image maybe that's why it is getting slower than others.

From next versioning we can change our CI to use macos-14 and 15 images

PR #152, opened 2025-09-17
CI: Adding and testing Big-Endian support for quaddtype

### LINKED ISSUE #119: Test little and big endian architectures in CI
We test little endian architectures in the numpy CI using PPC64le QEMU emulation. It's slow but it at least lets you do a real end-to-end test.

_Originally posted by @ngoldbaum in https://github.com/numpy/numpy-user-dtypes/pull/118#issuecomment-3094996347_

SwayamInSync:
closes #119 
I took inspiration from NumPy's workflow for this https://github.com/numpy/numpy/blob/main/.github/workflows/linux_qemu.yml
Since little-endian are already available in our current CI so here only big-endian testing.

Things seem to work and one test case is failing. Making the PR because CI is seems to work as expected


SwayamInSync:
Self note:
Interesting so endianness detection macors are working, somehow bit representation is not leading to the correct number or might be some other platform dependent intereference, will need to check further for this

But the main purpose of this PR is CI and that is working

juntyr:
Thanks @SwayamInSync for your work on this!

ngoldbaum:
Thanks @SwayamInSync!

PR #154, opened 2025-09-17
TYP: Static typing support for `numpy_quaddtype`

jorenham:
I heard that `numpy_quaddtype` has been growing lately, so I figured that it might be nice to have some static typing support. So here it is.

Mypy, stubtest, and pyright are all happy with it now (no errors are reported), even when they're run in strict mode.

To check for yourself, you can do something like 

```bash
uv venv .venv
uv pip install mypy pyright numpy .
source .venv/bin/activate
```

To run stubtest (validates correctness of stubs by inspecting the runtime, and also runs mypy to type-check the stubs):

```bash
$ stubtest --mypy-config-file pyproject.toml numpy_quaddtype
Success: no issues found in 2 modules
```

To run type-check the stubs using pyright (in strict mode):

```bash
$ pyright
0 errors, 0 warnings, 0 informations
```

If you want to admire the juicy type coverage %:

```bash
$ pyright --ignoreexternal --verifytypes numpy_quaddtype
# output omitted since that would spoil the fun ;)
```


quaddtype/numpy_quaddtype/_quaddtype_main.pyi
```diff
@@ -0,0 +1,134 @@
...
+    @override
+    def isnative(self) -> Literal[True]: ...
+    @property
+    @override
+    def isalignedstruct(self) -> Literal[False]: ...
+    @override
+    def __getitem__(self, key: Never, /) -> Self: ...  # type: ignore[override]
+
+@final
+class QuadPrecision:  # NOTE: It doesn't inherit from `np.generic` which is type-unsafe
```
jorenham, line 68: To elaborate a bit on this comment:
For all other dtypes (except `StringDType`, which doesn't even have a scalar type), the `dtype.type` is some subclass of `np.generic`. That means that it *should* be safe to do e.g. `arr.dtype.type.dtype`, since `np.generic.dtype`. But for `QuadPrecDType().dtype` there exists no `.type`.
This is called *type-unsafe* because there are situation where type-checkers are not able to detect these cases that will cause a `AttributeError` to be raised at runtime. 
This is of course not limited to `.dtype`, and there are (many) more examples that could go wrong like this.

SwayamInSync: In an ongoing work we subclassing the scalar to `PyFloatingArrType_Type` i.e. dtype.type can be a subclass of `np.generic`

```
In [4]: issubclass(npq.QuadPrecDType().type, np.generic)
Out[4]: True
```

This is currently waiting for one PR https://github.com/numpy/numpy/pull/29763 to get merged in NumPy and then we add this change here as well

SwayamInSync: does this resolve the type-unsafety?

jorenham: I take it that `PyFloatingArrType_Type` means that it'll subclass `np.floating` then?
Either way, it'll indeed be resolved then :)

jorenham: When it comes to that, and you need any help updating the stubs, I'd be happy to :)

SwayamInSync:
LGTM

quaddtype/numpy_quaddtype/_quaddtype_main.pyi
jorenham: > I've really no idea how you generated them :)

By writing them myself haha

quaddtype/numpy_quaddtype/_quaddtype_main.pyi
```diff
@@ -0,0 +1,134 @@
...
+    def __ge__(self, other: _IntoQuad, /) -> bool: ...
+
+    # Binary operators
+    def __add__(self, other: _CastsQuad, /) -> Self: ...
+    def __radd__(self, other: _CastsQuad, /) -> Self: ...  # type: ignore[misc]
+    def __sub__(self, other: _CastsQuad, /) -> Self: ...
+    def __rsub__(self, other: _CastsQuad, /) -> Self: ...
+    def __mul__(self, other: _CastsQuad, /) -> Self: ...
+    def __rmul__(self, other: _CastsQuad, /) -> Self: ...  # type: ignore[misc]
+    def __pow__(self, other: _CastsQuad, mod: None = None, /) -> Self: ...
```
juntyr, line 89: Where does the mod argument come from?

quaddtype/numpy_quaddtype/_quaddtype_main.pyi
```diff
@@ -0,0 +1,134 @@
...
+
+    # String representation
+    @override
+    def __repr__(self, /) -> str: ...
+    @override
+    def __str__(self, /) -> str: ...
+
+#
+def is_longdouble_128() -> bool: ...
+def get_sleef_constant(
```
juntyr, line 112: This is wrong - we need an override for the function since it returns normal integers for the integer constants bits and precision

quaddtype/numpy_quaddtype/_quaddtype_main.pyi
```diff
@@ -0,0 +1,134 @@
...
+        "bits",
+        "precision",
+        "resolution",
+    ],
+    /,
+) -> QuadPrecision: ...
+def set_num_threads(num_threads: int, /) -> None: ...
+def get_num_threads() -> int: ...
+def get_quadblas_version() -> Literal[
+    "QuadBLAS 1.0.0 - High Performance Quad Precision BLAS"
```
juntyr, line 133: I don't think this literal should be included

juntyr (commented):
We should also have a CI step to verify these type stubs

juntyr:
@jorenham @SwayamInSync I left some comments that should be addressed in a follow-up

quaddtype/numpy_quaddtype/_quaddtype_main.pyi
juntyr: Interesting, the operator __pow__ doesn't have it. You learn something new every day :)

quaddtype/numpy_quaddtype/_quaddtype_main.pyi
jorenham: Ah yea of course; `"bits"` and `"precision"` return an `int`. I'll fix this in a follow-up.

And out of curiosity, why is this a function instead of "bare" constants?

quaddtype/numpy_quaddtype/_quaddtype_main.pyi
juntyr: I'd prefer just leaving it as `def get_quadblas_version() -> str` - we promise to return *some* string but no other guarantees are made

quaddtype/numpy_quaddtype/_quaddtype_main.pyi
jorenham: Constants can be re-exported 🤷🏻

PR #159, opened 2025-09-19
`np.finfo` support for quaddtyoe

SwayamInSync:
Making this PR here as well, as a demonstration to how `np.finfo` and other related information can be passed to NumPy.
This maybe done from the Python side as well i.e. inside `__init__.py` something like
```python
QuadPrecDType.finfo = lambda: QuadPrecFinfo()
```
But this just feels so illegal that's why registering as `.tp_methods`

From the NumPy side it requires https://github.com/numpy/numpy/pull/29771 to get merged.

### DIFF
--- a/quaddtype/numpy_quaddtype/__init__.py
+++ b/quaddtype/numpy_quaddtype/__init__.py
@@ -8,6 +8,8 @@ from ._quaddtype_main import (
     get_quadblas_version
 )
 
+from dataclasses import dataclass
+
 __all__ = [
     'QuadPrecision', 'QuadPrecDType', 'SleefQuadPrecision', 'LongDoubleQuadPrecision',
     'SleefQuadPrecDType', 'LongDoubleQuadPrecDType', 'is_longdouble_128', 
@@ -15,7 +17,9 @@ __all__ = [
     'pi', 'e', 'log2e', 'log10e', 'ln2', 'ln10', 'max_value', 'epsilon',
     'smallest_normal', 'smallest_subnormal', 'bits', 'precision', 'resolution',
     # QuadBLAS related functions
-    'set_num_threads', 'get_num_threads', 'get_quadblas_version'
+    'set_num_threads', 'get_num_threads', 'get_quadblas_version',
+    # finfo class
+    'QuadPrecFinfo'
 ]
 
 def SleefQuadPrecision(value):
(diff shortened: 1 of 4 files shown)


SwayamInSync:
@juntyr can you please take a look at the `QuadPrecFinfo` class and see whether the info is correct?

quaddtype/numpy_quaddtype/src/dtype.c
```diff
@@ -224,13 +224,42 @@ QuadPrecDType_str(QuadPrecDTypeObject *self)
     return PyUnicode_FromFormat("QuadPrecDType(backend='%s')", backend_str);
 }
 
+static PyObject *
+QuadPrecDType_finfo(QuadPrecDTypeObject *self, PyObject *args)
+{
+    PyObject *numpy_quaddtype_module = PyImport_ImportModule("numpy_quaddtype");
```
SwayamInSync, line 230: I don't like doing this as well, but for now it works so

quaddtype/numpy_quaddtype/__init__.py
```diff
@@ -43,3 +47,39 @@ def LongDoubleQuadPrecDType():
...
+    """Floating-point information for quadruple precision dtype.
+    
+    This class provides information about the floating-point representation
+    used by the QuadPrecDType, similar to numpy.finfo but customized for
+    quad precision arithmetic.
+    """
+    bits: int = int(bits)
+    eps: float = float(epsilon)
+    epsneg: float = float(epsilon)
+    iexp: int = int(precision)
```
juntyr, line 62: iexp should be 15, I think?

quaddtype/numpy_quaddtype/__init__.py
```diff
@@ -43,3 +47,39 @@ def LongDoubleQuadPrecDType():
...
+    machar: object = None
+    machep: float = float(epsilon)
+    max: float = float(max_value)
+    maxexp: float = float(max_value)
```
juntyr, line 66: This would probably be int(ceil(max_value))?

quaddtype/numpy_quaddtype/__init__.py
```diff
@@ -43,3 +47,39 @@ def LongDoubleQuadPrecDType():
...
+    min: float = float(smallest_normal)
+    minexp: float = float(smallest_normal)
+    negep: float = float(epsilon)
```
juntyr, line 69: int(log2(epsneg))

quaddtype/numpy_quaddtype/__init__.py
```diff
@@ -43,3 +47,39 @@ def LongDoubleQuadPrecDType():
...
+    nexp: int = int(bits) - int(precision) - 1
```
juntyr, line 70: maybe 15 + 1 (+1)? not sure

quaddtype/numpy_quaddtype/__init__.py
```diff
@@ -43,3 +47,39 @@ def LongDoubleQuadPrecDType():
...
+    nmant: int = int(precision)
```
juntyr, line 71: 113 or 112

quaddtype/numpy_quaddtype/__init__.py
```diff
@@ -43,3 +47,39 @@ def LongDoubleQuadPrecDType():
...
+@dataclass
+class QuadPrecFinfo:
+    """Floating-point information for quadruple precision dtype.
+    
+    This class provides information about the floating-point representation
+    used by the QuadPrecDType, similar to numpy.finfo but customized for
+    quad precision arithmetic.
+    """
+    bits: int = int(bits)
+    eps: float = float(epsilon)
```
juntyr, line 60: I am really uncomfortable with all the casts to float, since at least for some of them they will undoubtedly over- or underflow. In my mind, the finfo values should have the same dtype as the type we're describing, but I know this would likely be a breaking change; but likely one we need since otherwise e.g. finfo(quad).max don't make any sense

SwayamInSync:
> the finfo values should have the same dtype as the type we're describing, but I know this would likely be a breaking change; but likely one we need since otherwise e.g. finfo(quad).max don't make any sense

Hey yes good point, just fixed that

PR #160, opened 2025-09-28
remove macros that override NumPy macros

mattip:
I started playing with the quadtype code and noticed lots of warnings. Many are from sleef, but a few are from redifining a numpy macro. This removes both the macros (one is unused) and instead uses `PyType_Check` directly.


charris:
Thanks Matti.

PR #162, opened 2025-09-29
Implementing `fexpr` ufunc

SwayamInSync:
This is needed for the some internal finfo calculation hence, putting it here

### DIFF
--- a/quaddtype/tests/test_quaddtype.py
+++ b/quaddtype/tests/test_quaddtype.py
@@ -595,6 +595,7 @@ def test_hyperbolic_functions(op, val):
     rtol = 1e-13 if abs(float_result) < 1e100 else 1e-10
     np.testing.assert_allclose(float(quad_result), float_result, rtol=rtol, atol=1e-15,
                                err_msg=f"Value mismatch for {op}({val})")
+    
     # Check sign for zero results
     if float_result == 0.0:
         assert np.signbit(float_result) == np.signbit(


quaddtype/tests/test_finfo.py
juntyr, line 1: Why the new empty file?

quaddtype/tests/test_quaddtype.py
```diff
@@ -592,11 +592,276 @@ def test_hyperbolic_functions(op, val):
 
     # For finite non-zero results
     # Use relative tolerance for exponential functions due to their rapid growth
-    rtol = 1e-13 if abs(float_result) < 1e100 else 1e-10
```
juntyr, line 595: Why was this removed?

quaddtype/numpy_quaddtype/src/dtype.c
```diff
@@ -184,7 +184,6 @@ static PyType_Slot QuadPrecDType_Slots[] = {
         {NPY_DT_setitem, &quadprec_setitem},
         {NPY_DT_getitem, &quadprec_getitem},
         {NPY_DT_default_descr, &quadprec_default_descr},
-        {NPY_DT_PyArray_ArrFuncs_dotfunc, NULL},
```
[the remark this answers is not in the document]
SwayamInSync: Its was just redundant, when very earlier we thought this might be the slot to support matmul :)

PR #172, opened 2025-10-14
FIX: Allow promotion of Python abstract dtype + fill implementation

SwayamInSync:
closes numpy/numpy-quaddtype#21 

## Copilot Summary

This pull request enhances the quad-precision dtype implementation for NumPy by improving type compatibility and array filling functionality. The main changes include better handling of Python numeric types for dtype promotion and discovery, and the addition of a specialized fill function for generating arithmetic progressions in quad-precision arrays.

**Improved dtype compatibility and type discovery:**

* Updated `common_dtype` to treat Python abstract dtypes (`PyLongDType`, `PyFloatDType`) as compatible with quad-precision, ensuring correct type promotion when mixing these types.
* Enhanced `quadprec_discover_descriptor_from_pyobject` to accept Python `int` and `float` objects, returning a default quad-precision descriptor, and improved error handling for unsupported types.

**New array fill functionality:**

* Added a `quadprec_fill` function to efficiently fill quad-precision arrays with arithmetic progressions, supporting both Sleef and long double backends.
* Registered the new fill function in the dtype's slot table, enabling NumPy to use it for array operations.

### DIFF
--- a/quaddtype/tests/test_quaddtype.py
+++ b/quaddtype/tests/test_quaddtype.py
@@ -591,3 +591,74 @@ def test_hyperbolic_functions(op, val):
     if float_result == 0.0:
         assert np.signbit(float_result) == np.signbit(
             quad_result), f"Zero sign mismatch for {op}({val})"
+
+
+class TestTypePomotionWithPythonAbstractTypes:
+    """Tests for common_dtype handling of Python abstract dtypes (PyLongDType, PyFloatDType)"""
+    
+    def test_promotion_with_python_int(self):
+        """Test that Python int promotes to QuadPrecDType"""
+        # Create array from Python int
+        arr = np.array([1, 2, 3], dtype=QuadPrecDType)
+        assert arr.dtype.name == "QuadPrecDType128"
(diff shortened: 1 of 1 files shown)


juntyr:
Could you provide a reference to the docs for the fill function?

SwayamInSync:
> Could you provide a reference to the docs for the fill function?

Its not same as the user facing `ndarray.fill()`.  (You can use that without this implementation and its docs available [here](https://numpy.org/doc/2.1/reference/generated/numpy.ndarray.fill.html))
The one implemented is register within dtype level as `PyArray_ArrFuncs` you can see the implementation for other dtype here
https://github.com/numpy/numpy/blob/46d892084f1f2ce4f6312ebbbab2d094cf216b16/numpy/_core/src/multiarray/arraytypes.c.src#L3748

Can expose the custom implementation using `NPY_DT_PyArray_ArrFuncs_fill` slot

SwayamInSync:
Can I get a review here @ngoldbaum @seberg

PR #180, opened 2025-10-17
FEAT: `fabs` ufunc implementation

SwayamInSync:
## Copilot Summary
This pull request adds support for the `np.fabs` universal function (ufunc) for the custom `QuadPrecision` dtype, ensuring it behaves identically to `np.absolute` for real numbers. The implementation includes both the underlying C++ registration and comprehensive tests for various input values. The release tracker documentation is also updated to reflect this new support.

**Feature addition:**

* Added the `fabs` ufunc for the `QuadPrecision` dtype in the C++ extension by registering it in `init_quad_unary_ops`. This ensures that `np.fabs` can be used with quad-precision numbers.

**Testing:**

* Introduced a new parameterized test `test_fabs` in `test_quaddtype.py` to verify the correctness of `np.fabs` for `QuadPrecision` scalars and arrays, covering positive, negative, zero, special, and small values. The test also checks that the sign of zero is positive after applying `fabs`.

**Documentation:**

* Updated `release_tracker.md` to mark `fabs` as supported for both the ufunc and its corresponding Python API, reflecting the new implementation status.

### DIFF
--- a/quaddtype/numpy_quaddtype/src/umath/unary_ops.cpp
+++ b/quaddtype/numpy_quaddtype/src/umath/unary_ops.cpp
@@ -156,6 +156,8 @@ init_quad_unary_ops(PyObject *numpy)
     if (create_quad_unary_ufunc<quad_absolute, ld_absolute>(numpy, "absolute") < 0) {
         return -1;
     }
+    // fabs is simialr to absolute, just not handles complex values (we neither)
+    // registering the absolute here
     if (create_quad_unary_ufunc<quad_absolute, ld_absolute>(numpy, "fabs") < 0) {
         return -1;
     }


SwayamInSync:
similar to absolute, just does not handle the complex types (so do we) hence registering absolute as fabs

quaddtype/numpy_quaddtype/src/umath/unary_ops.cpp
```diff
@@ -156,6 +156,9 @@ init_quad_unary_ops(PyObject *numpy)
     if (create_quad_unary_ufunc<quad_absolute, ld_absolute>(numpy, "absolute") < 0) {
         return -1;
     }
+    if (create_quad_unary_ufunc<quad_absolute, ld_absolute>(numpy, "fabs") < 0) {
```
juntyr: Could you add a comment for why we alias - that we don't handle complex values - just in case we would support that in the future?

SwayamInSync:
Let's take this in as well!

PR #182, opened 2025-10-17
FEAT: `conj` and `conjugate` ufunc implementation

SwayamInSync:
## Copilot Summary
This pull request adds support for the `conjugate` and `conj` unary operations for quad and long double types, treating them as no-ops since these types represent real numbers. It also updates documentation to reflect this new support.

### New unary operations support

* Added `quad_conjugate` and `ld_conjugate` functions to `ops.hpp`, implementing conjugate as a no-op for real numbers. [[1]](diffhunk://#diff-ad4140fa03d374f246bdc4d441d868fa759788f2d27a7a29ca259b16780f0850R40-R46) [[2]](diffhunk://#diff-ad4140fa03d374f246bdc4d441d868fa759788f2d27a7a29ca259b16780f0850R227-R233)
* Registered the `conjugate` unary operation in `init_quad_unary_ops` and documented that `conj` is an alias, so no separate registration is needed.

### Documentation update

* Updated `release_tracker.md` to mark `conj` and `conjugate` as supported for both quad and long double types.

### DIFF
--- a/quaddtype/tests/test_quaddtype.py
+++ b/quaddtype/tests/test_quaddtype.py
@@ -1693,3 +1693,25 @@ def test_heaviside_broadcast():
     assert result.dtype.name == "QuadPrecDType128"
     np.testing.assert_array_equal(result.astype(float), expected)
 
+
+@pytest.mark.parametrize("func", [np.conj, np.conjugate])
+@pytest.mark.parametrize("value", [
+    0.0,
+    -0.0,
+    1.5,
+    -1.5,
+    np.inf,
+    -np.inf,
+    np.nan,
+])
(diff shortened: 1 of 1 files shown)


SwayamInSync:
> Could this PR have a very basic no-op test (what do we do for np.positive)?

for np.positive and others we just compare against the numpy's answer on float64 with some relative tolerance

SwayamInSync:
One reason I didn't add the tests because this covers only the real part of conjugate functionality i.e only returning the value back like identity.

juntyr:
I still think it would be good to test this, otherwise one could implement the function by returning pi

SwayamInSync:
Coool tests are added, lets merge this as well

PR #188, opened 2025-10-18
FEAT: All logical ufunc implementation

SwayamInSync:
## Copilot Summary
This pull request adds support for logical operations (`logical_and`, `logical_or`, `logical_xor`, and `logical_not`) for the QuadPrecision data type, ensuring their behavior matches NumPy's semantics. The implementation covers both Sleef quad and long double backends, integrates these operations into the ufunc system, and provides comprehensive tests to verify correctness and compatibility.

**Logical operations implementation:**

* Added helper functions and logical operation implementations for both `Sleef_quad` and `long double` backends in `ops.hpp`, including truthiness checks and all four logical operations (`and`, `or`, `xor`, `not`).

**Ufunc integration:**

* Registered new binary logical ufuncs (`logical_and`, `logical_or`, `logical_xor`) in `umath/comparison_ops.cpp`.
* Implemented and registered the unary logical not ufunc with correct output type resolution and strided loop handling in `umath/unary_ops.cpp`. [[1]](diffhunk://#diff-b2cc0421155e8dbdd5001448fa86223d912cdac4ee1ea6b3d9a302bff539297cR147-R269) [[2]](diffhunk://#diff-b2cc0421155e8dbdd5001448fa86223d912cdac4ee1ea6b3d9a302bff539297cR388-R393)

**Testing and documentation:**

* Added thorough tests for all logical operations, including edge cases (zero, negative zero, NaN, inf), in `test_quaddtype.py`, comparing QuadPrecision results against NumPy's float64 behavior.
* Updated `release_tracker.md` to mark logical operations as implemented and tested.

### DIFF
--- a/quaddtype/numpy_quaddtype/src/ops.hpp
+++ b/quaddtype/numpy_quaddtype/src/ops.hpp
@@ -1247,7 +1247,7 @@ quad_is_nonzero(const Sleef_quad *a)
 {
     // A value is falsy if it's exactly zero (positive or negative)
     // NaN and inf are truthy
-    npy_bool is_zero = Sleef_icmpeqq1_purec(*a, QUAD_ZERO);
+    npy_bool is_zero = Sleef_icmpeqq1(*a, QUAD_ZERO);
     return !is_zero;
 }
 


quaddtype/numpy_quaddtype/src/ops.hpp
```diff
@@ -1238,3 +1238,83 @@ ld_greaterequal(const long double *a, const long double *b)
...
+
+// Logical operations
+
+// Helper function to check if a Sleef_quad value is non-zero (truthy)
+static inline npy_bool
+quad_is_nonzero(const Sleef_quad *a)
+{
+    // A value is falsy if it's exactly zero (positive or negative)
+    // NaN and inf are truthy
+    npy_bool is_zero = Sleef_icmpeqq1_purec(*a, QUAD_ZERO);
```
juntyr: Why are we using the purec version here, I think we've never used the specific prefixes?

juntyr (approved):
LGTM, with the one comment addressed

quaddtype/numpy_quaddtype/src/ops.hpp
SwayamInSync: Ah yes they are pure c implementation is just a specialization and "no-suffix" one can sometimes map to FMA versions if hardware supports else pure-c.

Let me change it to keep it consistent

SwayamInSync:
Merging this in!

PR #190, opened 2025-10-19
FEAT: `nextafter` ufunc implementation

SwayamInSync:
This PR adds the `nextafter` ufunc support to quaddtype. I referenced the [gcc's libquadmath's implementation for __float128](https://github.com/gcc-mirror/gcc/blob/master/libquadmath/math/nextafterq.c) and tried to map it back in SLEEF.

For testing we can't do the normal value comparison against numpy with other dtype like float64 so I added checking the properties like
- nextafter(1,2) => eps
- nextafter(0, 1) => subnormal
- nextafter(+inf, finite) => max_finfite
and similar more checking symmetry and direction. I also tested the C implementation against the libquadmath and it seems to be passing there as well.

### DIFF
--- a/quaddtype/numpy_quaddtype/src/ops.hpp
+++ b/quaddtype/numpy_quaddtype/src/ops.hpp
@@ -1003,8 +1003,8 @@ quad_nextafter(const Sleef_quad *x, const Sleef_quad *y)
         return result;
     }
     
-    // well I did not read to have those above checks
-    // but they can be important when settng FPE flag manually
+    // well I did not need to have those above checks
+    // but they can be important when setting FPE flag manually
     return quad_set_words64(hx, lx);
 }
 


juntyr:
Spelling error (but for some reason I cannot comment inline right now): "did not need the checks" (at the final sleef return)

juntyr:
Once we also have an implementation of the spacing ufunc, both ufuncs should have a test using the other ufunc

SwayamInSync:
> What does nextafter(-0.0, 1.0) and nextafter(+0.0, -1.0) do?

nextafter(-0, 1) => (+) smallest subnormal
nextafter(+0, -1) => (-) subnormal (can't call it smallest as logically it is the biggest negative value representable)

Its like 1st argument defines where you are and 2nd in which direction (not value) you want to go and returns the very next representable floating value

> Spelling error (but for some reason I cannot comment inline right now): "did not need the checks" (at the final sleef return)

Ah sorry, will fix that

SwayamInSync:
> Once we also have an implementation of the spacing ufunc, both ufuncs should have a test using the other ufunc

For `spacing` I was thinking we can just do
```
nextafter(x, inf) - x
```

PR #197, opened 2025-10-23
BUG: Fix Round-tripping QuadPrecision through string 

SwayamInSync:
closes numpy/numpy-quaddtype#18 

Cause:
For conversion from quad -> string, earlier we were using the strict 33 as significant decimal digit precision for dragon4 (which is the minimum value) so some random inputs were getting successfully round-trip but not all.
For example with original behaviour, `np.random.seed(0)` should pass but not `np.random.seed(1)` or others

128-bit quad support 33-36 so we should be setting this value to the max, as for any irregular representation Quad(x) can give upto max 36 significant decimal digit precision but our conversion to string will only preserve 33.

source: https://en.wikipedia.org/wiki/Quadruple-precision_floating-point_format

### DIFF
--- a/quaddtype/numpy_quaddtype/src/scalar.c
+++ b/quaddtype/numpy_quaddtype/src/scalar.c
@@ -15,6 +15,12 @@
 #include "scalar_ops.h"
 #include "dragon4.h"
 
+// For IEEE 754 binary128 (quad precision), we need 36 decimal digits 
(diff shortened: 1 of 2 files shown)


juntyr (approved):
LGTM!

ngoldbaum:
LGTM!

PR #199, opened 2025-10-23
FEAT: Extending `QuadPrecision` input to support arrays, sequences and np.numbers

SwayamInSync:
closes numpy/numpy-quaddtype#17

PR #201, opened 2025-10-24
TYP: Allow passing ``floating``, ``integer``, and ``bool_`` scalars to ``QuadPrecision.__new__``

jorenham:
This follows #199. See the added `# NOTE`  for why we cannot annotate the array-like constructor cases.

~I didn't see any mention of `np.bool_` scalars in #199, only floating and integer ones. So I take it that `QuadPrecision.__new__` does not accept `np.bool_` instances then?~


quaddtype/numpy_quaddtype/_quaddtype_main.pyi
```diff
@@ -64,8 +64,15 @@ class QuadPrecDType(np.dtype[QuadPrecision]):  # type: ignore[misc, type-var]  #
     @override
     def __getitem__(self, key: Never, /) -> Self: ...  # type: ignore[override]
 
+# NOTE: Until `QuadPrecision` will become a subclass of `np.generic`, this class cannot
```
SwayamInSync:
```
In [5]: issubclass(npq.QuadPrecDType().type, np.generic)
Out[5]: True

In [6]: npq.QuadPrecDType().type
Out[6]: numpy_quaddtype.QuadPrecision

In [7]: issubclass(npq.QuadPrecision, np.generic)
Out[7]: True
```

I think it is as Quadprecision already subclassing the np.floating

jorenham: Nice! I'll open a separate PR for that then.

SwayamInSync:
>  So I take it that QuadPrecision.__new__ does not accept np.bool_ instances then?

I think we should accept this right :) @jorenham I'll do the PR for this, you can assume it accept

cc: @ngoldbaum

SwayamInSync:
Also for the part that QuadPrecision is type-safe (it subclassing np.generic)

jorenham:
Yea this is fine to merge. I want to address the np.generic subclass stuff in a follow-up, since that will probably be a bit more involved

SwayamInSync:
Cool thanks @jorenham merging this in

PR #206, opened 2025-10-24
configure dependabot for github actions

jorenham:
I noticed several outdated action versions in the workflows, so I figured that dependabot could help with that.


SwayamInSync:
Thanks @jorenham merging this in

PR #208, opened 2025-10-25
FEAT: `imag` and `real` getters for QuadPrecision

SwayamInSync:
closes numpy/numpy-quaddtype#16


SwayamInSync:
Nice, yeah make sense, will be happy to extend this in NumPy whenever you feel right!

ngoldbaum:
One thing that might help for ensuring new code has tests is to look at integrating coverage testing. That way you would have gotten a test failure here because the coverage went down.

SwayamInSync:
> One thing that might help for ensuring new code has tests is to look at integrating coverage testing. That way you would have gotten a test failure here because the coverage went down.

I know coverage.py does something like this (never used it though) is that what you recommending?

PR #211, opened 2025-10-26
TEST: Trignometric ufunc tests

SwayamInSync:
closes numpy/numpy-quaddtype#32 

Here since for large inputs, argument reduction can lead to a very different values than `float64` hence comparing the quadprecision agains `mpmath` with `prec=113` instead of NumPy.

There are some cases where mpmath does not follow the IEEE standards, those cases are specifically handled

### DIFF
--- a/quaddtype/tests/test_quaddtype.py
+++ b/quaddtype/tests/test_quaddtype.py
@@ -1771,7 +1771,8 @@ class TestTrignometricFunctions:
     quad_result = numpy_op(quad_val)
     mpmath_result = mpmath_op(mpf_val)
     # convert mpmath result to quad for comparison
-    mpmath_result = QuadPrecision(str(mpmath_result))
+    # Use mp.nstr to get full precision (40 digits for quad precision)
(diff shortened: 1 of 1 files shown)


quaddtype/numpy_quaddtype/_quaddtype_main.pyi
```diff
@@ -81,6 +81,12 @@ class QuadPrecision:
     # See https://github.com/python/mypy/issues/18343#issuecomment-2571784915
     def __new__(cls, /, value: _IntoQuad, backend: _Backend = "sleef") -> Self: ...
 
+    # Attributes
```
SwayamInSync: These new attributes aren't added in stubs for the PR #208 , small addition do did in this PR

juntyr:
Once we merge this PR, it would be good to issue another release, e.g. v0.3, so that it is easier to test the new functionality and find any other issues that we want to resolve before a v1.0 release.

@ngoldbaum When will numpy v2.4 be released, which (correct me if I'm wrong) we now depend on for the improved dtype and finfo support?

SwayamInSync:
Merging this in!

PR #213, opened 2025-10-27
ENH: Support arbitrary-length Python ints in QuadPrecision constructor

SwayamInSync:
This PR is a part of work, where making `numpy_quaddtype `compatible with NumPy's longdouble tests

## Details
Use `PyLong_AsLongLongAndOverflow` instead of `PyLong_AsLongLong`, and handle the overflow case explicitly by reconstructing the `QuadPrecision` object via string conversion.


ngoldbaum (approved):
Thanks!

PR #221, opened 2025-10-31
FEAT: Implementing `is_integer` and `as_integer_ratio` for `QuadPrecision`

SwayamInSync:
closes numpy/numpy-quaddtype#14

I took the reference for implementing `as_integer_ratio` from [CPython's implementation](https://github.com/python/cpython/blob/ac1ffd77858b62d169a08040c08aa5de26e145ac/Objects/floatobject.c#L1503C1-L1572C2)

PR #222, opened 2025-10-31
CI: fix inclusion path in typecheck workflow

jorenham:
I noticed that it wasn't running in #221. This should fix it.


SwayamInSync:
Thanks @jorenham taken this in

PR #225, opened 2025-11-13
FEAT: Adding array casting support to fixed length strings

SwayamInSync:
This PR introduces the support for `np.str_` and `str` (will work on bytes after this gets merged)

### DIFF
--- a/quaddtype/numpy_quaddtype/src/casts.cpp
+++ b/quaddtype/numpy_quaddtype/src/casts.cpp
@@ -181,12 +181,21 @@ unicode_to_quad_resolve_descriptors(PyObject *NPY_UNUSED(self), PyArray_DTypeMet
                                     PyArray_Descr *given_descrs[2], PyArray_Descr *loop_descrs[2],
                                     npy_intp *view_offset)
 {
-    Py_INCREF(given_descrs[0]);
-    loop_descrs[0] = given_descrs[0];
+    if (!PyArray_ISNBO(given_descrs[0]->byteorder)) {
+        loop_descrs[0] = PyArray_DescrNewByteorder(given_descrs[0], NPY_NATIVE);
+        if (loop_descrs[0] == nullptr) {
+            return (NPY_CASTING)-1;
+        }
+    }
+    else {
+        Py_INCREF(given_descrs[0]);
+        loop_descrs[0] = given_descrs[0];
+    }
 
     if (given_descrs[1] == NULL) {
         loop_descrs[1] = (PyArray_Descr *)new_quaddtype_instance(BACKEND_SLEEF);
         if (loop_descrs[1] == nullptr) {
+            Py_DECREF(loop_descrs[0]);
             return (NPY_CASTING)-1;
         }
     }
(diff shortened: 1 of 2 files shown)


quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -151,38 +157,407 @@ quad_to_quad_strided_loop_aligned(PyArrayMethod_Context *context, char *const da
...
+        npy_intp i;
+        for (i = 0; i < copy_len; i++) {
+            Py_UCS4 c = ucs4_str[i];
+
+            // reject non-ASCII characters
+            if (c > 127) {
+                PyErr_Format(PyExc_ValueError,
+                             "Cannot cast non-ASCII character '%c' to QuadPrecision", c);
+                return -1;
+            }
```
ngoldbaum: won't the string-conversion functions called by `cstring_to_quad` catch this?

By the way, you might take inspiration from the `string_to_long_double` function in numpy:

https://github.com/numpy/numpy/blob/fabf1844667c81344ec0d0ab4958cde9cb9a845c/numpy/_core/src/multiarray/arraytypes.c.src#L533-L595

You may also want to simply copy/paste and vendor the code in `NumPyOS_ascii_strtold` into this project to get a parser that act exactly like numpy's.

ngoldbaum (commented):
Just a few comments. I didn't carefully review the reference counting.

quaddtype/numpy_quaddtype/src/casts.cpp
SwayamInSync: Hmm reading the code, there is not much difference, we directly working with UCS but NumPy converts it first to UTF-8 ( we can do this) and special case handling, white spaces, etc are already being handled by the `Sleef_strotoq` and `strtold` except the handling of `locale` which we might not be able to do as 
- SLEEF's conversion function does no support that
- `strtold_l` (used by `NumPyOS_ascii_strtold`)  is not C standard (a posix extension)

SwayamInSync: https://github.com/numpy/numpy/blob/fabf1844667c81344ec0d0ab4958cde9cb9a845c/numpy/_core/src/multiarray/arraytypes.c.src#L579

this line is the issue

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -151,38 +157,391 @@ quad_to_quad_strided_loop_aligned(PyArrayMethod_Context *context, char *const da
...
 }
 
+// Unicode/String to QuadDType casting
+static NPY_CASTING
+unicode_to_quad_resolve_descriptors(PyObject *NPY_UNUSED(self), PyArray_DTypeMeta *dtypes[2],
+                                    PyArray_Descr *given_descrs[2], PyArray_Descr *loop_descrs[2],
+                                    npy_intp *view_offset)
+{
+    if (!PyArray_ISNBO(given_descrs[0]->byteorder)) {
+        loop_descrs[0] = PyArray_DescrNewByteorder(given_descrs[0], NPY_NATIVE);
```
SwayamInSync: Creating new one might be better than just throwing incompatibility error

PR #228, opened 2025-11-20
FEAT: Adding Quad to and from bytes array casting support

SwayamInSync:
Similar implementation to #225 (only endianness does not apply here)

### DIFF
--- a/quaddtype/numpy_quaddtype/src/casts.cpp
+++ b/quaddtype/numpy_quaddtype/src/casts.cpp
@@ -342,7 +342,8 @@ quad_to_string_adaptive(Sleef_quad *sleef_val, npy_intp unicode_size_chars)
         return NULL;
     }
 
-    npy_intp pos_len = strlen(pos_str);
+    // no need to scan full, only checking if its longer
+    npy_intp pos_len = strnlen(pos_str, unicode_size_chars + 1);
 
     // If positional format fits, use it; otherwise use scientific notation
     if (pos_len <= unicode_size_chars) {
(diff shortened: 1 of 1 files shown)


SwayamInSync (commented):
Sorry it took me 3 days, I was learning more about the compiler optimizations and here is the [godbolt compiler explorer](https://godbolt.org/z/59zq3G5rx) link to see for proof

ngoldbaum:
If there are any spots you'd particularly like review for that would help. This is a big diff!

ngoldbaum:
I've been buried under notifications all week and haven't had time. I'm planning to look this over today though.

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -483,38 +507,107 @@ quad_to_unicode_loop_aligned(PyArrayMethod_Context *context, char *const data[],
...
+    // If target descriptor is wide enough, it's a safe cast
+    if (loop_descrs[1]->elsize >= required_size_bytes) {
+        return NPY_SAFE_CASTING;
+    }
+    return NPY_SAME_KIND_CASTING;
+}
+
+// Helper function: Copy string to bytes output buffer
+static inline void
+copy_string_to_bytes(const char *str, char *out_bytes, npy_intp bytes_size)
```
ngoldbaum: I'd call this `copy_cstring_to_bytes` to make it clearer that `str` has to be null-terminated.

ngoldbaum: or just inline it because the function only has one caller

SwayamInSync:
> LGTM now.
> 
> It occurs to me that adding a test run that builds CPython, NumPy, and quaddtype using address sanitizer and then runs the tests. There's an ASan test run in the NumPy CI if you want an example. Keep in mind that you need to build NumPy with `pip` instead of `spin` if you use that example though.

Make sense, I'll check and extend the suite this weekend

Issue #231, opened 2025-11-20
Issues with hashing/saving QuadPrecDType


dlakaplan:
Within PINT we currently need to be able to hash individual QuadPrecision values (for data integrity checks) and save arrays to `npz` files for rapid access.  I don't think either of those currently work:

```
import numpy as np
from numpy_quaddtype import QuadPrecDType, QuadPrecision

hash(QuadPrecision(np.float32(2)))
```
fails (while `hash(np.float128(2))` works using the native `numpy` type).  

And:
```
d = np.array([1, 2, 3], dtype=QuadPrecDType)
np.savez('test.npz',d=d)
```
fails while the same for `d=np.array([1,2,3],dtype=np.float128)` works.  

My hope is that these are both relatively easy fixes?

ngoldbaum:
I initially added pickle support in https://github.com/numpy/numpy-user-dtypes/pull/33/files. One load-bearing piece is to add `NPY_LIST_PICKLE` to the dtype flags: https://github.com/numpy/numpy-user-dtypes/commit/457fc7671eea91acbfe479a389bd5bae4bcc2e48

You can see the latest version of the implementation in numpy by looking at `stringdtype__reduce__`: https://github.com/numpy/numpy/blob/dd80e20bc4b7f78ba92177c87d11c236bb7d194d/numpy/_core/src/multiarray/stringdtype/dtype.c#L725-L757

dlakaplan:
Hi, @SwayamInSync   Actually, I was looking for updates on this issue (not the one I mentioned - apologies).  Are there updates here I could test?

SwayamInSync:
Yeah so here the pickling/unpickling of arrays support is added.
Hashing of scalar is todo, will wrap that as well very soon

PR #232, opened 2025-11-22
FEAT: Adding pickle support to quaddtype

SwayamInSync:
This PR adds the pickle support as per numpy/numpy-quaddtype#9 
Usage:

```python
original = np.array([1.5, 2.5, 3.5, 4.5], dtype=QuadPrecDType(backend=backend))
# Pickle and unpickle
pickled = pickle.dumps(original)
unpickled = pickle.loads(pickled)
```

Using `np.savez`

```python
arr1 = np.array([1.5, 2.5, 3.5], dtype=QuadPrecDType(backend=backend))
arr2 = np.array([[1.0, 2.0], [3.0, 4.0]], dtype=QuadPrecDType(backend=backend))

np.savez(fname, array1=arr1, array2=arr2)
loaded = np.load(fname, allow_pickle=True)	# allow_pickle must be True during loading
loaded_arr1 = loaded['array1']
loaded_arr2 = loaded['array2']
```


SwayamInSync:
hash implementation might require some careful handling https://github.com/python/cpython/blob/20b69aac0d19a5e5358362410d9710887762f0e7/Python/pyhash.c#L87

I'll do this in a different PR and then we can close the #231

juntyr:
Why is the backend provided as an integer? A string literal might be a lot clearer, or if it has to be an int maybe we could add an int enum?

SwayamInSync:
> Why is the backend provided as an integer? A string literal might be a lot clearer, or if it has to be an int maybe we could add an int enum?

Internally `backend `is defined as enum, we provide the friendly API for the user to add input as string. For the sake of static tools, IDE suggestions I think defining as enum might make sense maybe @jorenham can give some inputs here

quaddtype/numpy_quaddtype/__init__.pyi
```diff
@@ -10,9 +11,15 @@ from ._quaddtype_main import (
     set_num_threads,
 )
 
+class QuadBackend(enum.IntEnum):
+    """Backend type for QuadPrecision computations."""
```
jorenham: docstrings aren't needed in stubs (there's even a `flake8-pyi` rule that prohibits it)

```suggestion
```

quaddtype/tests/test_quaddtype.py
```diff
@@ -4292,4 +4292,462 @@ def test_as_integer_ratio_compatibility_with_float(self, value):
...
+        pickled = pickle.dumps(original)
+        unpickled = pickle.loads(pickled)
+        
+        # Verify array is preserved
+        np.testing.assert_array_equal(unpickled, original)
+        assert unpickled.dtype == original.dtype
+        assert unpickled.flags.f_contiguous == original.flags.f_contiguous
+    
+    @pytest.mark.parametrize("backend", ["sleef", "longdouble"])
+    def test_pickle_reduce_method(self, backend):
```
ngoldbaum: My only comment is that this test is a little too tightly coupled to the implementation to be useful.

You could, for example, decide to use `__getnewargs__` instead of `__reduce__` but then you'd have to update this test too. https://docs.python.org/3/library/pickle.html#object.__getnewargs__.

SwayamInSync:
Merging this in

PR #234, opened 2025-11-22
Fixing `QuadPrecision` scalar dtype exposure

SwayamInSync:
closes numpy/numpy-quaddtype#8

### DIFF
--- a/quaddtype/tests/test_quaddtype.py
+++ b/quaddtype/tests/test_quaddtype.py
@@ -4293,3 +4293,18 @@ class Test_Is_Integer_Methods:
         quad_ratio = quad_num / quad_denom
         float_ratio = float_num / float_denom
         assert abs(quad_ratio - float_ratio) < 1e-15
+
+def test_quadprecision_scalar_dtype_expose():
+    quad_ld = QuadPrecision("1e100", backend="longdouble")
+    quad_sleef = QuadPrecision("1e100", backend="sleef")
+    assert quad_ld.dtype == QuadPrecDType(backend='longdouble')
+    assert np.dtype(quad_ld) == QuadPrecDType(backend='longdouble')
+    
+    #todo: Uncomment them when 232 is merged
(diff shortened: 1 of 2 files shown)


quaddtype/tests/test_quaddtype.py
```diff
@@ -4292,4 +4292,19 @@ def test_as_integer_ratio_compatibility_with_float(self, value):
...
         float_ratio = float_num / float_denom
-        assert abs(quad_ratio - float_ratio) < 1e-15
\ No newline at end of file
+        assert abs(quad_ratio - float_ratio) < 1e-15
+
+def test_quadprecision_scalar_dtype_expose():
+    quad_ld = QuadPrecision("1e100", backend="longdouble")
+    quad_sleef = QuadPrecision("1e100", backend="sleef")
+    assert quad_ld.dtype == QuadPrecDType(backend='longdouble')
+    assert np.dtype(quad_ld) == QuadPrecDType(backend='longdouble')
```
seberg: Of course `scalar.dtype` should work and I guess since you are parametric you have to make it work.

Butt the protocol is the exact opposite of this assert.  It specifically wants to *disable* this in the future because it makes no sense to say `np.ones(..., dtype=scalar)`, the user should write `dtype=scalar.dtype`.

So in other words: The attribute is right, the `__numpy_dtype__` protocol is just unrelated.

seberg: What we could try to have is a function to map from the scalar type to the DType class cleanly, I guess (this exists in C, but not Python).
I used to want `np.dtype[scalar_type]` for this, but while what typing does is conceptually the same, it's not compatible :(.

seberg:
> hold off on `__numpy_dtype__` - no code in the real world uses it yet.

Just be overly clear onc emore: It isn't even correct to implement it! Scalars should not be convertible to dtypes via that mechanism.

PR #240, opened 2025-12-06
FIX: SLEEF migration to 4.0

SwayamInSync:
closes numpy/numpy-quaddtype#5 

Saturday spend well! Some points to mention:
- String parsing helpers inside `utilities.c` and their corresponding calls have been modified because of the modified behaviour of `Sleef_strtoq` and handling of negative NaN.
- extra dependency of `tlfloat`
- revision is set to the latest commit upto the date of this PR (`43a0252`), not setting to `master` as maybe in future more routine behaviour changes can break the dtype here (Ideally in future it'll be better to migrate the core changes from master if something fundamental breaks)
- SSE Instructions are by default disabled in this new version (maybe maintainig them is cost effective) ``QBLAS ``still uses them, I'll update them there and then we can maybe update here or keep it as it is
- Added more tests that replicate the bug from the mentioned issue, and similar pattern for other binary ops

### DIFF
--- a/quaddtype/numpy_quaddtype/src/ops.hpp
+++ b/quaddtype/numpy_quaddtype/src/ops.hpp
@@ -13,13 +13,13 @@ typedef Sleef_quad (*unary_op_quad_def)(const Sleef_quad *);
 // Unary Quad operations with 2 outputs (for modf, frexp)
 typedef void (*unary_op_2out_quad_def)(const Sleef_quad *, Sleef_quad *, Sleef_quad *);
 
-[[maybe_unused]] static Sleef_quad
+static inline Sleef_quad
 quad_negative(const Sleef_quad *op)
 {
     return Sleef_negq1(*op);
 }
 
-[[maybe_unused]] static Sleef_quad
+static inline Sleef_quad
 quad_positive(const Sleef_quad *op)
 {
     return *op;


juntyr (approved):
LGTM

SwayamInSync:
Took the opportunity to also clean the building process and warnings

SwayamInSync:
So tl;dr I refactored the string parsing logic and now every call site is using `NumPyOS_ascii_strtoq` (which first parses the input, handling inf/nan, leading whitespaces) and then if the quantitty is still parsable then delegates to **internal helper** `cstring_to_quad_internal` (this is not exposed and it just convert the remaining entity to the quad) 

SLEEF 4.0's `strtoq`'s all limitations are handled by this internal helper so in future if they fix those we just need to update this internal helper, the exposed public API will remain the same

cc: @ngoldbaum I hope you like this setup

quaddtype/numpy_quaddtype/src/ops.hpp
```diff
@@ -13,13 +13,13 @@ typedef Sleef_quad (*unary_op_quad_def)(const Sleef_quad *);
...
 
-static Sleef_quad
+[[maybe_unused]] static Sleef_quad
 quad_negative(const Sleef_quad *op)
 {
     return Sleef_negq1(*op);
 }
 
-static Sleef_quad
+[[maybe_unused]] static Sleef_quad
```
seberg: Why not make it static inline like all others then?

SwayamInSync: Done

PR #242, opened 2025-12-08
Compiler Sanitization tests CI

SwayamInSync:
@ngoldbaum from NumPy I added both ASAN and UBSAN, for now only UBSAN warnings seems to be there.
The workflow is set to fail only if any warnings are found. Maybe we can more loose it to as fail if only ASAN are found.


juntyr:
@SwayamInSync Will this PR be revived in some other form before or after 1.0?

PR #244, opened 2025-12-19
FEAT: Adding variable length `StringDType` to/from QuadDType array casting.

SwayamInSync:
As per the title

closes numpy/numpy-quaddtype#11

### DIFF
--- a/quaddtype/numpy_quaddtype/src/casts.cpp
+++ b/quaddtype/numpy_quaddtype/src/casts.cpp
@@ -360,13 +360,11 @@ quad_to_string_adaptive(Sleef_quad *sleef_val, npy_intp unicode_size_chars)
     if (pos_len <= unicode_size_chars) {
         return positional_str;  // Keep the positional string
     }
-    else {
-        Py_DECREF(positional_str);
-        // Use scientific notation with full precision
-        return Dragon4_Scientific_QuadDType(sleef_val, DigitMode_Unique,
-                                           SLEEF_QUAD_DECIMAL_DIG, 0, 1,
-                                           TrimMode_LeaveOneZero, 1, 2);
-    }
+    Py_DECREF(positional_str);
+    // Use scientific notation with full precision
+    return Dragon4_Scientific_QuadDType(sleef_val, DigitMode_Unique,
+                                        SLEEF_QUAD_DECIMAL_DIG, 0, 1,
+                                        TrimMode_LeaveOneZero, 1, 2);
 }
 
 static inline const char *


SwayamInSync:
Tests contribute the major part of diff

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -605,6 +605,175 @@ quad_to_bytes_loop(PyArrayMethod_Context *context, char *const data[],
...
+            }
+            else {
+                NpyString_release_allocator(allocator);
+                PyErr_SetString(PyExc_ValueError, "Cannot convert null string to QuadPrecision");
+                return -1;
+            }
+        }
+
+        quad_value out_val;
+        if (bytes_to_quad_convert(s.buf, s.size, backend, &out_val) < 0) {
```
SwayamInSync: utilising `bytes_to_quad_convert` instead of `unicode_to_quad_convert` because the later expects `Py_UCS` format otherwise both are doing the same thing

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -605,6 +605,175 @@ quad_to_bytes_loop(PyArrayMethod_Context *context, char *const data[],
...
 }
 
+// StringDType to QuadDType casting
+static NPY_CASTING
+stringdtype_to_quad_resolve_descriptors(PyObject *NPY_UNUSED(self), PyArray_DTypeMeta *dtypes[2],
+                                        PyArray_Descr *given_descrs[2], PyArray_Descr *loop_descrs[2],
+                                        npy_intp *view_offset)
+{
+    Py_INCREF(given_descrs[0]);
+    loop_descrs[0] = given_descrs[0];
```
ngoldbaum: If you do this after checking `given_descrs[1]` then there's no need to decref in the error paths below so it'll be a little clearer

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -605,6 +605,175 @@ quad_to_bytes_loop(PyArrayMethod_Context *context, char *const data[],
...
+
+    npy_string_allocator *allocator = NpyString_acquire_allocator(str_descr);
+
+    while (N--) {
+        quad_value in_val = load_quad<Aligned>(in_ptr, backend);
+        Sleef_quad sleef_val = quad_to_sleef_quad(&in_val, backend);
+
+        // Get string representation with adaptive notation
+        // Use a large buffer size to allow for full precision
+        PyObject *py_str = quad_to_string_adaptive(&sleef_val, QUAD_STR_WIDTH);
```
ngoldbaum: There's no need to create a PyUnicode object here. Just pass the ASCII bytes of the C parsed string to `NpyString_Pack`.

ngoldbaum (commented):
I spotted some minor issues and one bigger opportunity to avoid unnecessarily creating python strings.

https://github.com/numpy/numpy-user-dtypes/pull/244/changes#r2636204832 is a bigger comment; don't consider it a blocker if you disagree or don't want to spend time on it.

quaddtype/numpy_quaddtype/src/casts.cpp
SwayamInSync: So actually this function `quad_to_string_adaptive` uses Dragon4 utilities like `Dragon4_Positional_QuadDType` and `Dragon4_Scientific_QuadDType` for conversion and they both returns a `PyUnicode_FromString` object and from them we extract the cstring from it.

I can modify the dragon4 helper or add one more helper with `_cstr` suffix that returns cstring and then this would be doable. Let me know if this sounds good?

SwayamInSync:
- `dragon4.c/.h` only contains some duplicate helpers that return `cstring` instead of PyUnicode object
- `casts.cpp` got another duplicate helper `quad_to_string_adaptive_cstr` that calls corresponding dragon4 utilities to return `cstring`
- tests are consolidated into others

PR #246, opened 2025-12-26
FEAT: Implementing `same_value` casting rule in quaddtype

SwayamInSync:
closes numpy/numpy-quaddtype#26 

As per the title

### DIFF
--- a/quaddtype/numpy_quaddtype/src/casts.cpp
+++ b/quaddtype/numpy_quaddtype/src/casts.cpp
@@ -30,7 +30,22 @@ extern "C" {
 #include "constants.hpp"
 
 #define NUM_CASTS 40  // 18 to_casts + 18 from_casts + 1 quad_to_quad + 1 void_to_quad
-#define QUAD_STR_WIDTH 50  // 42 is enough for scientific notation float128, just keeping some buffer
+
+/*
+For quad precision scientific notation, we need at most:
+
+1 character for sign
+1 character for leading digit
+1 character for decimal point
+36 significant digits
+1 character for e
+1 character for exponent sign
+4 characters for exponent (max is 4932)
+1 null terminator
+
+Total: 46 characters, using 50 as a safe buffer
+*/
+#define QUAD_STR_WIDTH 50
 
 // forward declarations
 static inline const char *


SwayamInSync:
## Details for easing the review process

Only `casts.cpp` and `test_quaddtype` requires the major review, I added comments wherever I made the changes to give reasoning for that step, the rest files are modified for the caues listed below

## Small refactoring
- `quad_value` union with `__float128` causes lot of ABI & compiler optimization issues in C++, hence refactoring the methods to not perform any kind of copy of this passed union instead use pointers
- Inter-backend operation is broken in previous builds, not sure why someone wants to go in that direction, but just for the sake of completeness I fixed it here as well, to implement the `same_value` casting between `QuadPrecision` with different backends
- Remove the aligned/unlaigned separate loops with their templated versions

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -26,88 +27,116 @@ extern "C" {
...
-        loop_descrs[1] = given_descrs[1];
-        if (given_descrs[0]->backend != given_descrs[1]->backend) {
-            casting = NPY_UNSAFE_CASTING;
+
+    Py_INCREF(given_descrs[1]);
+    loop_descrs[1] = given_descrs[1];
+
+    if (given_descrs[0]->backend != given_descrs[1]->backend) {
+        // Different backends require actual conversion, no view possible
+        *view_offset = NPY_MIN_INTP;
```
juntyr: what does the view_offset mean?

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -26,88 +27,116 @@ extern "C" {
...
-        Py_INCREF(given_descrs[1]);
-        loop_descrs[1] = given_descrs[1];
-        if (given_descrs[0]->backend != given_descrs[1]->backend) {
-            casting = NPY_UNSAFE_CASTING;
+
+    Py_INCREF(given_descrs[1]);
+    loop_descrs[1] = given_descrs[1];
+
+    if (given_descrs[0]->backend != given_descrs[1]->backend) {
+        // Different backends require actual conversion, no view possible
```
juntyr: I thought we stored them in a union so that views would be possible? Do we know why not?

In any case explicit casting is better here since it better communicates that loss may happen.

I know we also export a function to check if long double is float128, if so should we allow the cast from sleef to long double as safe?

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -26,88 +27,116 @@ extern "C" {
...
+            Sleef_quad temp = Sleef_cast_from_doubleq1(static_cast<double>(ld));
+            memcpy(&roundtrip.sleef_value, &temp, sizeof(Sleef_quad));
+        }
+        
+        // Compare in SLEEF domain
+        if (Sleef_iunordq1(in_val->sleef_value, roundtrip.sleef_value))
+            return 1;  // Both NaN
+        if (Sleef_icmpeqq1(in_val->sleef_value, roundtrip.sleef_value))
+            return 1;  // Equal
+        if (Sleef_icmpeqq1(in_val->sleef_value, QUAD_ZERO) && Sleef_icmpeqq1(roundtrip.sleef_value, QUAD_ZERO))
```
juntyr: Why do we have this extra check? If the previous equality check doesn't catch +0.0 == -0.0, then this one would still check (+0.0 == +0.0) && (-0.0 == +0.0) which is no better

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -417,37 +497,38 @@ quad_to_unicode_loop(PyArrayMethod_Context *context, char *const data[],
     QuadBackendType backend = descr_in->backend;
 
     npy_intp unicode_size_chars = descrs[1]->elsize / 4;
+    int same_value_casting = ((context->flags & NPY_SAME_VALUE_CONTEXT_FLAG) == NPY_SAME_VALUE_CONTEXT_FLAG);
```
juntyr: same as above

quaddtype/numpy_quaddtype/src/scalar_ops.cpp
```diff
@@ -134,6 +134,9 @@ quad_richcompare(QuadPrecisionObject *self, PyObject *other, int cmp_op)
         Py_INCREF(other);
         other_quad = (QuadPrecisionObject *)other;
         if (other_quad->backend != backend) {
+            // we could allow, but this will be bad
+            // Two values that are different in quad precision, 
+            // might appear equal when converted to double.
```
juntyr:
```suggestion
            // we could allow this, but it would be bad
            // since two values that are different in quad precision
            // might appear equal when converted to double.
```

quaddtype/tests/test_quaddtype.py
```diff
@@ -5368,3 +5368,296 @@ def test_hash_backends(self, backend):
...
+    
+        for val in failing:
+            q = np.array([val], dtype=QuadPrecDType())
+            with pytest.raises(ValueError):
+                q.astype(dtype, casting="same_value")
+
+    @pytest.mark.parametrize("dtype", [
+        np.float16, np.float32, np.float64, np.longdouble
+    ])
+    @pytest.mark.parametrize("val", [0.0, -0.0, float('inf'), float('-inf'), float('nan')])
```
juntyr: please add -nan and check that the sign is preserved, same for +0.0 and -0.0

quaddtype/tests/test_quaddtype.py
```diff
@@ -5368,3 +5368,296 @@ def test_hash_backends(self, backend):
...
+
+    @pytest.mark.parametrize("dtype", [
+        "S50", "U50", "S100", "U100", np.dtypes.StringDType()
+    ])
+    def test_same_value_cast_strings_enough_width(self, dtype):
+        """Test that string types with enough width can represent quad values exactly."""
+        values = [
+            "0.0", "-0.0", "1.0", "-1.0",
+            "3.14159265358979323846264338327950288",  # pi with full quad precision
+            "inf", "-inf", "nan",
```
juntyr: what happens for -nan

quaddtype/tests/test_quaddtype.py
```diff
@@ -5368,3 +5368,296 @@ def test_hash_backends(self, backend):
...
+            if np.isnan(q[0]):
+                assert np.isnan(back[0]), f"NaN roundtrip failed for {dtype}"
+            else:
+                assert q[0] == back[0], f"Value {val} roundtrip failed for {dtype}"
+
+    @pytest.mark.parametrize("dtype", ["S10", "U10"])
+    def test_same_value_cast_strings_narrow_width(self, dtype):
+        """Test that string types with narrow width fail for values that need more precision."""
+        # Values that can fit in 10 chars should pass
+        passing_values = ["0.0", "1.0", "-1.0", "inf", "-inf", "nan"]
```
juntyr: also check -0.0 and -nan

quaddtype/numpy_quaddtype/src/casts.cpp
SwayamInSync: Union is a convenient and efficient way to store only one of the values, that does not gurantee that stored bytes will be interpreted in same manner (atleast on systems where __float128 is not defined and longdouble != 128 bits)
On systems `Sleef_quad` can be `__float128` or a struct of 2 `int64`, the interpretation and handling of bytes will be needed for cross-platform support.

> I know we also export a function to check if long double is float128, if so should we allow the cast from sleef to long double as safe?

This is indeed possible but I believe this flag should represent a general scenario, as if somebody on their system has `longdouble == 128 bits` and they checked and found "oh cast is safe" and they thought its universal as in NumPy we try to keep things supported cross-platform. Although I am against of users performing inter-backend operations.

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -26,88 +27,116 @@ extern "C" {
...
+        else if (std::isinf(ld)) {
+            roundtrip.sleef_value = (ld > 0) ? QUAD_PRECISION_INF : QUAD_PRECISION_NINF;
+        }
+        else {
+            Sleef_quad temp = Sleef_cast_from_doubleq1(static_cast<double>(ld));
+            memcpy(&roundtrip.sleef_value, &temp, sizeof(Sleef_quad));
+        }
+        
+        // Compare in SLEEF domain
+        if (Sleef_iunordq1(in_val->sleef_value, roundtrip.sleef_value))
```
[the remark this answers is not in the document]
SwayamInSync: Right I forgot this is an `OR` condition, I'll fix this to assess on based on inputs AND ensuring NAN roundtrip

quaddtype/numpy_quaddtype/src/utilities.c
```diff
@@ -262,9 +262,13 @@ Sleef_quad
...
-        return in_val->sleef_value;
+        // can directly return but that causes union heisenbugs, 
+        // but this helper is rare to use, so acceptable
+        Sleef_quad result;
+        memcpy(&result, &in_val->sleef_value, sizeof(Sleef_quad));
+        return result;
     }
     else {
-        return Sleef_cast_from_doubleq1(in_val->longdouble_value);
+        return Sleef_cast_from_doubleq1((double)(in_val->longdouble_value));
```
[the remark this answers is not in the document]
SwayamInSync: static_cast is C++ can't use in .c files, but here its safe

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -117,44 +146,63 @@ quad_to_quad_strided_loop_aligned(PyArrayMethod_Context *context, char *const da
 
     QuadPrecDTypeObject *descr_in = (QuadPrecDTypeObject *)context->descriptors[0];
     QuadPrecDTypeObject *descr_out = (QuadPrecDTypeObject *)context->descriptors[1];
+    QuadBackendType backend_in = descr_in->backend;
+    QuadBackendType backend_out = descr_out->backend;
+    int same_value_casting = ((context->flags & NPY_SAME_VALUE_CONTEXT_FLAG) == NPY_SAME_VALUE_CONTEXT_FLAG);
```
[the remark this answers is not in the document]
SwayamInSync: can do, but that's very low priority, if everything else sounds good

SwayamInSync:
### New things fixed and added
- For supporting `-nan` in strings, dragon4 has to preserve the sign in output to ensure the complete roundtrip
- Sleef's vendored methods for casting to/from double aren't all supportable over platforms, can cause discripencies so we are explicitly handling those cases.

quaddtype/numpy_quaddtype/src/ops.hpp
```diff
@@ -1,12 +1,10 @@
 #include <sleef.h>
 #include <sleefquad.h>
 #include <cmath>
+#include "constants.hpp"
 
 // Quad Constants, generated with qutil
-#define QUAD_ZERO sleef_q(+0x0000000000000LL, 0x0000000000000000ULL, -16383)
 #define QUAD_ONE sleef_q(+0x1000000000000LL, 0x0000000000000000ULL, 0)
```
juntyr: we can also get rid of this one

quaddtype/numpy_quaddtype/src/ops.hpp
```diff
@@ -649,12 +647,12 @@ quad_mod(const Sleef_quad *a, const Sleef_quad *b)
 
     // Handle zero result sign: when result is exactly zero,
     // it should have the same sign as the divisor (NumPy convention)
-    if (Sleef_icmpeqq1(result, QUAD_ZERO)) {
-        if (Sleef_icmpltq1(*b, QUAD_ZERO)) {
-            return Sleef_negq1(QUAD_ZERO);  // -0.0
+    if (Sleef_icmpeqq1(result, QUAD_PRECISION_ZERO)) {
+        if (Sleef_icmpltq1(*b, QUAD_PRECISION_ZERO)) {
+            return Sleef_negq1(QUAD_PRECISION_ZERO);  // -0.0
```
juntyr: we have a const for negative zero

quaddtype/tests/test_quaddtype.py
```diff
@@ -5454,11 +5454,13 @@ def test_same_value_cast_quad_to_int(self, dtype, passing, failing):
...
         np.float16, np.float32, np.float64, np.longdouble
     ])
-    @pytest.mark.parametrize("val", [0.0, -0.0, float('inf'), float('-inf'), float('nan')])
+    @pytest.mark.parametrize("val", [0.0, -0.0, float('inf'), float('-inf'), float('nan'), float("-nan")])
     def test_same_value_cast_floats_special_values(self, dtype, val):
         """Test that special floating-point values roundtrip correctly."""
         q = np.array([val], dtype=QuadPrecDType())
         result = q.astype(dtype, casting="same_value")
+        if str(val).startswith("-"):
+            assert np.signbit(result), f"Sign bit failed for {dtype} with value {val}"
```
juntyr: could we assert `np.signbit(result) == str(val).startswith("-")` instead?

quaddtype/tests/test_quaddtype.py
```diff
@@ -5601,7 +5607,9 @@ def test_quad_to_quad_same_value_casting_passing(self, src_backend, dst_backend)
             result = src.astype(QuadPrecDType(backend=dst_backend), casting="same_value")
             
             # Verify value is preserved
-            if val == "nan":
+            if str(val).startswith("-"):
```
juntyr: same as above

ngoldbaum (commented):
@mattip any chance you have time and interest to look this over?

SwayamInSync:
> ## Details for easing the review process
> Only `casts.cpp` and `test_quaddtype` requires the major review, I added comments wherever I made the changes to give reasoning for that step, the rest files are modified for the caues listed below
> 
> ## Small refactoring
> * `quad_value` union with `__float128` causes lot of ABI & compiler optimization issues in C++, hence refactoring the methods to not perform any kind of copy of this passed union instead use pointers
> * Inter-backend operation is broken in previous builds, not sure why someone wants to go in that direction, but just for the sake of completeness I fixed it here as well, to implement the `same_value` casting between `QuadPrecision` with different backends
> * Remove the aligned/unlaigned separate loops with their templated versions

This is still valid, outside this are just tweaks and refactoring to make things work in stable manner

ngoldbaum:
@SwayamInSync do you mind waiting a couple weeks on this? I'd like to have Matti look because he added `same_value` casting and also has a try at this in #161.

In the meantime: one thing that would help is to make the PR more atomic. There's a fair amount of renaming and other stuff like that you could do in another PR then rebase this once the other PR is merged to reduce the diff size.

In general I'd much rather review several smaller more focused PRs rather than one big PR.

juntyr:
I'm also hoping that we can merge this PR and publish v1.0 soon

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -1310,6 +1392,56 @@ from_quad<long double>(const quad_value *x, QuadBackendType backend)
...
+    }
+    Sleef_quad sleef_val = quad_to_sleef_quad(x, backend);
+    const char *val_str = quad_to_string_adaptive_cstr(&sleef_val, QUAD_STR_WIDTH);
+    if (val_str != NULL) {
+        PyErr_Format(PyExc_ValueError, 
+                     "QuadPrecision value '%s' cannot be represented exactly in the target dtype",
+                     val_str);
+    }
+    else {
+        PyErr_SetString(PyExc_ValueError,
```
ngoldbaum: same error

mattip (approved):
LGTM, pending the already-mentioned corrections. Tests should be catching edge cases.

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -752,6 +859,14 @@ quad_to_stringdtype_strided_loop(PyArrayMethod_Context *context, char *const dat
 
         Py_ssize_t str_size = strnlen(str_buf, QUAD_STR_WIDTH);
 
+        // Perform same_value check if requested
+        if (same_value_casting) {
+            if (quad_to_string_same_value_check(&in_val, str_buf, str_size, backend) < 0) {
```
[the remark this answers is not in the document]
SwayamInSync: Oh yes right, my bad
we shouldn't be needing same_value check here.

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -371,6 +414,65 @@ quad_to_string_adaptive_cstr(Sleef_quad *sleef_val, npy_intp unicode_size_chars)
...
+    Sleef_quad sleef_val = quad_to_sleef_quad(in_val, backend);
+    const char *val_str = quad_to_string_adaptive_cstr(&sleef_val, QUAD_STR_WIDTH);
+    if (val_str != NULL) {
+        PyErr_Format(PyExc_ValueError,
+                     "QuadPrecision value '%s' cannot be represented exactly in target string dtype "
+                     "(string width too narrow or precision loss occurred)",
+                     val_str);
+    }
+    else {
+        PyErr_SetString(PyExc_ValueError,
```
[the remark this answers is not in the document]
SwayamInSync: Fixed

quaddtype/numpy_quaddtype/src/casts.cpp
SwayamInSync: Fixed

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -371,6 +414,65 @@ quad_to_string_adaptive_cstr(Sleef_quad *sleef_val, npy_intp unicode_size_chars)
 
 }
 
+static inline int
+quad_to_string_same_value_check(const quad_value *in_val, const char *str_buf, npy_intp str_len,
```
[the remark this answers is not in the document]
SwayamInSync: There's a nuance here, let me explain so 36 for exact decimal digits and 42 for scientific notation
They are not exposed by SLEEF, we keep the max string len to 50 (42 + some buffer)

Now when casting a QuadPrecision value to a fixed-width string (like U10) then the dragon4 will generate full buffer representation but it will get truncated to only 10 chars. Then here we in same_value check we need to re-create the null-terminated buffer. 

Oh good point, that buffer can be in local stack, we don't be needing the heap allocation here.

ngoldbaum (approved):
Thanks for your patience here!

SwayamInSync:
Great! Merging this in.
Thanks everyone 🙌

juntyr:
Thanks! How long do you think will it now take to release v1.0?

PR #247, opened 2026-01-02
Refactor: General refactors + quad2quad inter-backend fixes

SwayamInSync:
This PR contains the unrelated refactors happend inside #246 (the commits were inter-twined with `same_value` feat so had to pick them manually here, can cause conflict in that PR but should be resolvable)
- Single function to handle Sleef to double casting with edge cases
- Use pass by pointer instead of value for `quad_value` union
- Fixed the Quad2Quad inter-backend casting

### DIFF
--- a/quaddtype/numpy_quaddtype/src/casts.cpp
+++ b/quaddtype/numpy_quaddtype/src/casts.cpp
@@ -950,11 +950,21 @@ inline quad_value
 to_quad<spec_npy_half>(npy_half x, QuadBackendType backend)
 {
     quad_value result;
+    double d = npy_half_to_double(x);
     if (backend == BACKEND_SLEEF) {
-        result.sleef_value = Sleef_cast_from_doubleq1(npy_half_to_double(x));
+        if (std::isnan(d)) {
+            result.sleef_value = std::signbit(d) ? QUAD_PRECISION_NEG_NAN : QUAD_PRECISION_NAN;
+        }
+        else if (std::isinf(d)) {
+            result.sleef_value = (d > 0) ? QUAD_PRECISION_INF : QUAD_PRECISION_NINF;
+        }
+        else {
+            Sleef_quad temp = Sleef_cast_from_doubleq1(d);
+            std::memcpy(&result.sleef_value, &temp, sizeof(Sleef_quad));
+        }
     }
     else {
-        result.longdouble_value = (long double)npy_half_to_double(x);
+        result.longdouble_value = (long double)d;
     }
     return result;
 }
@@ -964,8 +974,18 @@ inline quad_value
 to_quad<float>(float x, QuadBackendType backend)
 {
     quad_value result;
-    if (backend == BACKEND_SLEEF) {
-        result.sleef_value = Sleef_cast_from_doubleq1(x);
+    if (backend == BACKEND_SLEEF) 
+    {
+      if (std::isnan(x)) {
+          result.sleef_value = std::signbit(x) ? QUAD_PRECISION_NEG_NAN : QUAD_PRECISION_NAN;
+      }
+      else if (std::isinf(x)) {
+          result.sleef_value = (x > 0) ? QUAD_PRECISION_INF : QUAD_PRECISION_NINF;
+      }
+      else {
+          Sleef_quad temp = Sleef_cast_from_doubleq1(static_cast<double>(x));
+          std::memcpy(&result.sleef_value, &temp, sizeof(Sleef_quad));
+      }
     }
     else {
         result.longdouble_value = (long double)x;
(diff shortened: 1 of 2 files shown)


SwayamInSync:
We could've also refactored the `to_quad` template specilizations to also take union pointers intead of returining, but that does not seems to cause any cross-platform issues (also compiler might perform the return value optimization, so doesn't seems like harming performance)

quaddtype/numpy_quaddtype/src/ops.hpp
```diff
@@ -649,12 +644,12 @@ quad_mod(const Sleef_quad *a, const Sleef_quad *b)
 
     // Handle zero result sign: when result is exactly zero,
     // it should have the same sign as the divisor (NumPy convention)
-    if (Sleef_icmpeqq1(result, QUAD_ZERO)) {
-        if (Sleef_icmpltq1(*b, QUAD_ZERO)) {
-            return Sleef_negq1(QUAD_ZERO);  // -0.0
+    if (Sleef_icmpeqq1(result, QUAD_PRECISION_ZERO)) {
+        if (Sleef_icmpltq1(*b, QUAD_PRECISION_ZERO)) {
+            return Sleef_negq1(QUAD_PRECISION_ZERO);  // -0.0
```
juntyr: we already have a negative zero constant

quaddtype/numpy_quaddtype/src/ops.hpp
```diff
@@ -618,8 +613,8 @@ static inline Sleef_quad
 quad_mod(const Sleef_quad *a, const Sleef_quad *b)
 {
     // division by zero
-    if (Sleef_icmpeqq1(*b, QUAD_ZERO)) {
-        return QUAD_NAN;
+    if (Sleef_icmpeqq1(*b, QUAD_PRECISION_ZERO)) {
+        return QUAD_PRECISION_NAN;
```
juntyr: do we have guarantees about the sign of the NaN here?

quaddtype/numpy_quaddtype/src/ops.hpp
```diff
@@ -670,13 +665,13 @@ quad_fmod(const Sleef_quad *a, const Sleef_quad *b)
     }
     
     // Division by zero -> NaN
-    if (Sleef_icmpeqq1(*b, QUAD_ZERO)) {
-        return QUAD_NAN;
+    if (Sleef_icmpeqq1(*b, QUAD_PRECISION_ZERO)) {
+        return QUAD_PRECISION_NAN;
```
juntyr: same here

quaddtype/numpy_quaddtype/src/utilities.c
```diff
@@ -262,9 +262,13 @@ Sleef_quad
...
-        return in_val->sleef_value;
+        // can directly return but that causes union heisenbugs, 
+        // but this helper is rare to use, so acceptable
+        Sleef_quad result;
+        memcpy(&result, &in_val->sleef_value, sizeof(Sleef_quad));
+        return result;
     }
     else {
-        return Sleef_cast_from_doubleq1(in_val->longdouble_value);
+        return Sleef_cast_from_doubleq1((double)(in_val->longdouble_value));
```
juntyr: can we use a static_cast here?

quaddtype/numpy_quaddtype/src/casts.cpp
```diff
@@ -1111,188 +1125,188 @@ numpy_to_quad_strided_loop_aligned(PyArrayMethod_Context *context, char *const d
...
 inline npy_byte
-from_quad<npy_byte>(quad_value x, QuadBackendType backend)
+from_quad<npy_byte>(const quad_value *x, QuadBackendType backend)
 {
     // runtime warnings often comes from/to casting of NaN, inf
     // casting is used by ops at several positions leading to warnings
     // fix can be catching the cases and returning corresponding type value without casting
     if (backend == BACKEND_SLEEF) {
-        return (npy_byte)Sleef_cast_to_int64q1(x.sleef_value);
+        return (npy_byte)Sleef_cast_to_int64q1(x->sleef_value);
```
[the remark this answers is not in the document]
SwayamInSync: Sure, here we can, although the casting coming from a int64 to lower variants so the current code is also safe

quaddtype/numpy_quaddtype/src/utilities.c
SwayamInSync: Here we can't as it's a C file

quaddtype/numpy_quaddtype/src/ops.hpp
SwayamInSync: That's just an artifact in the code, I shouldn't have added that but just a future todo :)

quaddtype/numpy_quaddtype/src/dragon4.c
```diff
@@ -974,22 +974,22 @@ PrintInfNan(char *buffer, npy_uint32 bufferSize, npy_uint64 mantissa, npy_uint32
...
+    if (signbit == '+') {
+        if (pos < maxPrintLen - 1) {
+            buffer[pos++] = '+';
+        }
+    }
+    else if (signbit == '-') {
+        if (pos < maxPrintLen - 1) {
+            buffer[pos++] = '-';
+        }
+    }
```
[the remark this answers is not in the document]
SwayamInSync: Maybe for the sake of completeness yeah, I am not sure of the actual usecase of -nan
An even important case is -0.0 as I've seen some people use it as signal and it is working correct in numpy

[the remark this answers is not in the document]
juntyr: -0.0 is, unfortunately, quite important, so that you can go from e.g. -1/inf=-0 back to -1/-0=inf.

-NaN is more niche but worth preserving nonetheless

PR #256, opened 2026-01-16
Expand API docs and add CHANGELOG

juntyr:
API docs and CHANGELOG for the docs, based on @SwayamInSync work. I reviewed the API docs carefully, including all numerical values, and used KaTex for better maths rendering. I also cleaned up the CHANGELOG a bit, though I'm less confident about that one (so I'd be happy to drop it for now). I briefly reviewed the docs index page but didn't edit it further, so follow-ups might want to adjust the cards based on what guides we write


juntyr:
Superseded by https://github.com/numpy/numpy-quaddtype/pull/49

